PR #2, opened 2020-12-24
MEN-4241: improve the handling of the stop message

tranchitella:


tranchitella:
Green pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/236791627

connection/connection.go
```diff
@@ -130,18 +132,68 @@ func NewConnection(u url.URL,
...
+	}
+
+	pingPeriod := (c.defaultPingWait * 9) / 10
+	ticker := time.NewTicker(pingPeriod)
+	defer ticker.Stop()
+
+	c.connection.SetPongHandler(func(string) error {
+		log.Debug("PongHandler called")
+		// requires go >= 1.15
+		// ticker.Reset(pingPeriod)
```
lluiscampos, line 158: golang 1.14 is the latest supported in yocto warrior (and we did the port of this version ourselves). With the current plan, we won't be able to use 1.15 features until April 2023.

tranchitella: This is not an issue, it's a (micro) optimisation and that's why it is commented.

app/daemon_test.go
```diff
@@ -52,6 +52,10 @@ var (
 	testData              string
 )
 
+func init() {
+	connectionmanager.SetDefaultPingWait(10 * time.Second)
```
kacf, line 56: Since the test is one single binary, this affects all unit tests in this package. Is that intended, or would it be better to add this to setup/teardown functions?

kacf (approved):
A suggestion and a question below, but nothing that blocks the PR, it is good as is too.

app/daemon_test.go
tranchitella: It is intended, because the almost all the unit tests expect a lower value for ping timeouts.
This applies to all the unit tests in the `app` package, indeed.

shell/exec.go
```diff
@@ -99,7 +99,7 @@ func (s *MenderShell) pipeStdout() {
 		}
 		raw := make([]byte, 255)
 		n, err := sr.Read(raw)
-		if err != nil {
+		if err != nil || !s.IsRunning() {
 			if !s.IsRunning() {
 				return
 			}
```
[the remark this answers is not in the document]
tranchitella: You are right, way better:
https://github.com/mendersoftware/mender-connect/pull/2/commits/33b293b44b2397e5c3ef58fe6004ffbf3a1c5641

PR #6, opened 2021-01-06
Get user's home directory and use it as cwd and HOME env var value

tranchitella:
Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -79,6 +79,7 @@ type MenderShellDaemon struct {
 	terminalHeight          uint16
 	uid                     uint64
 	gid                     uint64
+	homeDir                 string
 	shellsSpawned           uint
 	debug                   bool
 }
@@ -359,6 +360,8 @@ func (d *MenderShellDaemon) Run() error {
 		return err
 	}
 
+	d.homeDir = u.HomeDir
+
 	d.uid, err = strconv.ParseUint(u.Uid, 10, 32)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


shell/shell.go
```diff
@@ -47,7 +48,12 @@ func ExecuteShell(uid uint32,
 		cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uid, Gid: gid}
 	}
 
+	if _, err := os.Stat(homeDir); !os.IsNotExist(err) {
+		cmd.Dir = homeDir
+	}
+	cmd.Env = append(cmd.Env, fmt.Sprintf("HOME=%s", homeDir))
```
kacf, line 54: Is it worth considering using something like `su -l <USER>` to make it a true login shell? I'm not sure exactly what all the differences are, but I know `su` does some extra things, like setting up the environment and putting the shell in its own process group. And specifically related to this PR, it sets the home directory.

tranchitella: @kacf can we expect to have `su -l user` everywhere, in all yocto environment et al?

tranchitella: Great, thanks for the confirmation.

tranchitella: @kacf one thing, though: if we do it this way, we have to run mender-connect as root, don't we?

app/daemon.go
```diff
@@ -372,6 +375,10 @@ func (d *MenderShellDaemon) Run() error {
 	log.Debug("mender-connect connecting to dbus")
 	//dbus main loop, required.
 	dbusAPI, err := dbus.GetDBusAPI()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
tranchitella: Done: https://github.com/mendersoftware/mender-connect/pull/6/commits/5e04d2991be66cded5d4b1cff53c0d2a46535b13

shell/shell.go
kacf: You're right. I thought it was possible to run it password-less if you su to the same user as the one running mender-connect, but this doesn't work. In that case I agree, better to stick to what we have for now.

kacf (approved):

PR #7, opened 2021-01-06
Home directory, stop message when the shell process exits

tranchitella:


shell/exec.go
```diff
@@ -102,8 +103,24 @@ func (s *MenderShell) pipeStdout() {
...
-			break
+			log.Errorf("error reading stdout: %s", err)
+			//
+			msg := &ws.ProtoMsg{
+				Header: ws.ProtoHdr{
+					Proto:     ws.ProtoTypeShell,
+					MsgType:   wsshell.MessageTypeStopShell,
+					SessionID: s.sessionId,
+					Properties: map[string]interface{}{
+						"status": wsshell.NormalMessage,
```
alfrunes, line 114: I'm not entirely sure about the `wsshell` specifics, but I believe it should be:
```suggestion
						"status": wsshell.ErrorMessage,
```

shell/exec.go
```diff
@@ -102,8 +103,24 @@ func (s *MenderShell) pipeStdout() {
...
+					},
+				},
+				Body: []byte(err.Error()),
+			}
+			err = connectionmanager.Write(ws.ProtoTypeShell, msg)
+			if err != nil {
+				log.Debugf("error on write: %s", err.Error())
+			}
+			return
 		} else if !s.IsRunning() {
```
alfrunes, line 124: What about this case (and before calling Read)? There is a potential race where the terminal can terminate outside blocking on `sr.Read`.

shell/exec.go
```diff
@@ -102,8 +103,24 @@ func (s *MenderShell) pipeStdout() {
 		raw := make([]byte, 255)
```
alfrunes, line 103: This might be optimized by the compiler, but can we move the buffer allocation outside the loop?

shell/exec.go
```diff
@@ -32,6 +32,8 @@ var (
 
 // MenderShellMessage represents a message between the device and the backend
 type MenderShellMessage struct {
```
alfrunes, line 34: This is starting to look suspiciously like `ws.ProtoMsg` :see_no_evil:

shell/exec.go
```diff
@@ -59,7 +61,6 @@ type MenderShell struct {
 }
 
 type MenderShellCommand struct {
```
alfrunes, line 63: I don't see this type being used anywhere, deadcode?

shell/exec.go
tranchitella: Technically, it is not an error but a "StopShell" command from the device.
We use ErrorMessage in response to a message, don't we?

shell/exec.go
tranchitella: If it terminates by other means (e.g. shutdown of the daemon), we manage it.
I believe the only case where we did not manage a clean shutdown is the termination of the process, which we detect with the inability to read its stdout. I may be wrong, though.

shell/exec.go
tranchitella: It predates ProtoMsg; we need to refactor the code to get rid of this struct, but as it is heavily work in progress, we are delaying it. I'll give it a try...

shell/exec.go
alfrunes: Yeah, naturally the file descriptors will close and return `io.EOF` (or `io.ErrUnexpectedEOF`?). Disregard...

shell/exec.go
alfrunes: TBH, I don't quite understand the semantics behind the "status" field. Is it required for all shell messages? @merlin-northern ?

shell/shell.go
```diff
@@ -47,7 +48,12 @@ func ExecuteShell(uid uint32,
 		cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uid, Gid: gid}
 	}
 
+	if _, err := os.Stat(homeDir); !os.IsNotExist(err) {
+		cmd.Dir = homeDir
+	}
```
merlin-northern, line 53: could we have some fallback in case user has no home directory, or the directory does not exists, e.g.: fall back to "/" ?

shell/shell_test.go
```diff
@@ -45,13 +45,13 @@ func TestMenderShellExecShell(t *testing.T) {
...
 	//command does not exist
-	pid, pseudoTTY, cmd, err := ExecuteShell(uint32(uid), uint32(gid), "thatissomethingthatdoesnotexecute", "xterm-256color", 24, 80)
+	pid, pseudoTTY, cmd, err := ExecuteShell(uint32(uid), uint32(gid), "/", "thatissomethingthatdoesnotexecute", "xterm-256color", 24, 80)
 	assert.Error(t, err)
 	assert.Equal(t, pid, -1)
 	assert.Nil(t, pseudoTTY)
 	assert.Nil(t, cmd)
 
-	pid, pseudoTTY, cmd, err = ExecuteShell(uint32(uid), uint32(gid), "/bin/sh", "xterm-256color", 24, 80)
+	pid, pseudoTTY, cmd, err = ExecuteShell(uint32(uid), uint32(gid), "/", "/bin/sh", "xterm-256color", 24, 80)
```
merlin-northern, line 54: could we have some explicit tests that it works? and that fallback works if the directory does not exist?
high-end test scenario: start terminal with given home, run the pwd command, compare the results. (copy paste from one of the tests that checks command execution)

shell/exec.go
merlin-northern: we are using status field as a status indicator, i.e.: if there is an error, we set "status" to indicate the error for the other side.

merlin-northern: it is not an error message, so status NormalMessage makes sense to me.
suggestion: can we make a helper function out of it, and provide unit tests for it? let's find a nice place and a nice name `SendMessage(status,body)`
just a suggestion.

shell/exec.go
```diff
@@ -92,12 +92,12 @@ func (s *MenderShell) IsRunning() bool {
 }
 
 func (s *MenderShell) pipeStdout() {
+	raw := make([]byte, 255)
```
merlin-northern, line 95: while you are at it, please define 255 as something.

merlin-northern (approved):
some comments. in general looks good. approving since I know you are going to make really good use of them ;-)

shell/shell.go
tranchitella: By default it uses the CWD of the mender-connect process, which is `/` in a "standard installation".
Do you mean hard-coded default, or a configurable one?

tranchitella: Let's do it with a hardcoded default to "/"

shell/exec.go
tranchitella: Done

shell/shell_test.go
tranchitella: Added a unit test

shell/exec.go
tranchitella: Here we get rid of it, in favour of ProtoMsg: https://github.com/mendersoftware/mender-connect/pull/7/commits/a33a9b70bd9b55586ea5c3ab97c5c4ff1d1d3a93

PR #9, opened 2021-01-08
MEN-4305: ping-pong based healthcheck for the remote terminal

tranchitella:


session/session.go
```diff
@@ -318,6 +327,10 @@ func (s *MenderShellSession) StartShell(sessionId string, terminal MenderShellTe
 	s.pseudoTTY = pseudoTTY
 	s.command = cmd
 	s.activeAt = timeNow()
+
+	// start the healthcheck go-routine
+	go s.healthcheck()
+
```
merlin-northern, line 333: I have some doubts if that go routine will be stopped in all cases when session is stopped. maybe we should make sure it is not left behind? (is it stopped on MenderSessionTerminateExpired?)

tranchitella: I believe so, it is managed by `StopShell` which closes the `close` chan. This causes the goroutine to exit.
MenderSessionTerminateExpired is calling StopShell, indeed.

app/daemon.go
```diff
@@ -709,6 +711,20 @@ func (d *MenderShellDaemon) routeMessageShellResize(message *ws.ProtoMsg) error
 	return nil
 }
 
+func (d *MenderShellDaemon) routeMessagePongShell(message *ws.ProtoMsg) error {
+	var err error
+
+	s := session.MenderShellSessionGetById(message.Header.SessionID)
+	if s == nil {
+		err = session.ErrSessionNotFound
+		d.routeMessageResponse(nil, err)
```
merlin-northern, line 720: @tranchitella this is the place wher npr happens (Alf recent report)

tranchitella: You are right, @merlin-northern
Fixed: https://github.com/mendersoftware/mender-connect/pull/9/commits/0a38b0a5277c273ff8a3c9d6bd132242abdd9dbd

session/session.go
```diff
@@ -65,6 +67,8 @@ var (
 	defaultSessionIdleExpiredTimeout = NoExpirationTimeout
 	defaultTimeFormat                = "Mon Jan 2 15:04:05 -0700 MST 2006"
 	MaxUserSessions                  = 1
+	healthcheckInterval              = time.Second * 60
```
merlin-northern, line 70: do you think we could maybe make it configurable? technically it will mean that we are sending data from the user device, maybe lets give the user ability to at least enable or disable keep alive and even change the interval?

session/session.go
```diff
@@ -341,6 +354,51 @@ func (s *MenderShellSession) IsExpired(setStatus bool) bool {
...
+	}
+	log.Debugf("session %s healthcheck ping", s.id)
+	err := connectionmanager.Write(ws.ProtoTypeShell, msg)
+	if err != nil {
+		log.Debugf("error on write: %s", err.Error())
+	}
+}
+
+func (s *MenderShellSession) HealthcheckPong() {
+	s.healthcheckTimeout = time.Now().Add(healthcheckInterval + healthcheckTimeout)
```
merlin-northern, line 399: this write happens in the daemon main loop; different thread than [this write](https://github.com/mendersoftware/mender-connect/pull/9/commits/17adb2019cc6028d858f364b9b05bde987599c4e#diff-10dab83c04a99358d431906d37f48b92be97d50a2719d4ddf81c62d3b03a0aebR359) and [these reads](https://github.com/mendersoftware/mender-connect/pull/9/commits/17adb2019cc6028d858f364b9b05bde987599c4e#diff-10dab83c04a99358d431906d37f48b92be97d50a2719d4ddf81c62d3b03a0aebR365) do you think we could use a mutex for protection?

Another thing (somewhat nitpicking): if I run a session on Mar 28th 2021 at 01:59 CEST or on Oct 31st 2021 at 02:59 CEST when DST start/end take place. will it mean my session will be in trouble?

merlin-northern (approved):
three comments, approving since I know you will make good use of them :)

session/session.go
kacf: > this write happens in the daemon main loop; different thread than [this write](https://github.com/mendersoftware/mender-connect/pull/9/commits/17adb2019cc6028d858f364b9b05bde987599c4e#diff-10dab83c04a99358d431906d37f48b92be97d50a2719d4ddf81c62d3b03a0aebR359) and [these reads](https://github.com/mendersoftware/mender-connect/pull/9/commits/17adb2019cc6028d858f364b9b05bde987599c4e#diff-10dab83c04a99358d431906d37f48b92be97d50a2719d4ddf81c62d3b03a0aebR365) do you think we could use a mutex for protection?

Well spotted. Due to the long timers between receiving a pong and triggering the next ping, triggering a race condition is extremely unlikely, but I still consider this somewhat bad style, and it may have more serious consequences later if this is changed. My suggested fix is: Instead of updating `s.healthcheckTimeout` here, use a channel to send the value back to the ping loop, and let it update the variable in there instead. Then all accesses are from the same thread.

kacf: > Another thing (somewhat nitpicking): if I run a session on Mar 28th 2021 at 01:59 CEST or on Oct 31st 2021 at 02:59 CEST when DST start/end take place. will it mean my session will be in trouble?

This should be fine, the Golang docs state that all "time differences" (as opposed to absolute times) are timezone independent.

kacf (changes requested):

session/session.go
tranchitella: Addressed here: https://github.com/mendersoftware/mender-connect/pull/9/commits/3d7bfc5baba385941b7c64df50adcced6ff899c9

tranchitella:
@kacf I addressed the concern you raised, can you please updated your review?

session/session.go
tranchitella: we could, but at this stage, a hard-coded 60 seconds looks "good enough". The clients (UI, mender-cli) will use the value mender-connect sends, thus we can easily make it configurable in the future if we need to.

kacf (approved):
Looks good, thanks!

PR #10, opened 2021-01-11
MEN-4339 Regression on removing the session in case process really exited

merlin-northern:
ChangeLog:none

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -633,7 +633,7 @@ func (d *MenderShellDaemon) routeMessageStopShell(message *ws.ProtoMsg) error {
 			}
 		}
 		d.routeMessageResponse(response, err)
-		return err
+		return session.MenderShellDeleteById(s.GetId())
 	}
 	if d.shellsSpawned == 0 {
 		log.Warn("can't decrement shellsSpawned count: it is 0.")


app/daemon.go
```diff
@@ -633,7 +633,7 @@ func (d *MenderShellDaemon) routeMessageStopShell(message *ws.ProtoMsg) error {
 			}
 		}
 		d.routeMessageResponse(response, err)
-		return err
+		return session.MenderShellDeleteById(s.GetId())
```
lluiscampos, line 636: I don't think this is correct, we need to return `err` regardless of `session.MenderShellDeleteById(s.GetId())` succession or not.

If any, we could wrap `err` with the potential error of `session.MenderShellDeleteById(s.GetId())`

app/daemon.go
```diff
@@ -633,7 +633,7 @@ func (d *MenderShellDaemon) routeMessageStopShell(message *ws.ProtoMsg) error {
 			}
```
lluiscampos, line 633: Also, line 626 above will `return` with no calling `session.MenderShellDeleteById(s.GetId())` -> won't this trigger the same kind of regression?

lluiscampos (changes requested):

tranchitella:
The proper fix is here: https://github.com/mendersoftware/mender-connect/pull/9/commits/7eb4c15225b52d2dfb9a7cece0e3bcb2b16499df
I'm closing this PR, we'll include the fix in this PR: https://github.com/mendersoftware/mender-connect/pull/9

PR #11, opened 2021-01-13
Allow to exit from mender-connect when waiting for the JWT token

tranchitella:
Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -14,6 +14,7 @@
 package app
 
 import (
+	"context"
 	"fmt"
 	"os/user"
 	"strconv"
@@ -57,12 +58,13 @@ type MenderShellDaemonEvent struct {
 }
 
 type MenderShellDaemon struct {
+	ctx                     context.Context
+	ctxCancel               context.CancelFunc
 	writeMutex              *sync.Mutex
 	eventChan               chan MenderShellDaemonEvent
 	connectionEstChan       chan MenderShellDaemonEvent
 	reconnectChan           chan MenderShellDaemonEvent
 	stop                    bool
-	stopChan                chan bool
 	authorized              bool
 	printStatus             bool
 	username                string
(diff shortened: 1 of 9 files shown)


app/daemon.go
```diff
@@ -402,7 +402,10 @@ func (d *MenderShellDaemon) Run() error {
 	}
 
 	jwtToken, err := client.GetJWTToken()
-	log.Debugf("GetJWTToken().len=%d,%v", len(jwtToken), err)
+	if err != nil {
+		return err
+	}
+	log.Debugf("GetJWTToken().len=%d", len(jwtToken))
```
kacf, line 422: There is an [explicit test](https://github.com/mendersoftware/integration/blob/master/tests/tests/test_mender_connect.py#L107-L121) which tests the opposite of this: It should start listening for the signal if the initial call to `GetJwtToken` fails. This is important because we do not have DBus activation enabled in the systemd service file, so there is no guarantee that the API is in fact up, even if the client is up. See [this commit](https://github.com/mendersoftware/mender/commit/3c868edbe815cc79cb5f8a67c5598c370ea1a043) for more information.

**Update**: While I was writing this you changed it. Are you on Slack? We should have a quick chat about it I think.

tranchitella: @kacf the failure here is because the mender-client is not up, thus the interface is not registered in the D-Bus system bus. The issue I'm facing is that listening on a remote object which does not exist (yet) doesn't work, you'll never get notifications. Thus, if you fail to call GetJWTToken, you cannot listen for signals. Let's discuss it online, indeed.

tranchitella: @kacf as agreed on Slack, I reworked the PR to not do anything magic: it just works fine if you start the Mender client *after* the Mender Connect service.

kacf (approved):
:+1:

tranchitella:
Manual pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/242635296

PR #15, opened 2021-02-06
Add remote port forwarding support to mender-connect

tranchitella:


config/config.go
```diff
@@ -72,6 +72,10 @@ type MenderShellConfigFromFile struct {
 	Sessions SessionsConfig `json:"Sessions"`
 	// Reconnect interval
 	ReconnectIntervalSeconds int
+	// Disable the Remote Terminal feature
+	DisableRemoteTerminal bool
+	// Disable the port forward feature
+	DisablePortForward bool
```
kacf, line 78: Double negatives. Would it be more natural to call them `EnablePortForward` and `EnableRemoteTerminal`? This also plays a bit better with the opt-in strategy we want to use for these features.

app/daemon_portforward.go
```diff
@@ -0,0 +1,227 @@
...
+
+	"github.com/mendersoftware/go-lib-micro/ws"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/mendersoftware/mender-connect/connectionmanager"
+)
+
+const (
+	ProtoTypePortForward       = 3
```
kacf, line 30: Maybe these top-level aliases should be grouped in a master file so we don't risk picking the same number? Git-merging could easily mess this up behind our backs.

app/daemon_portforward.go
```diff
@@ -0,0 +1,227 @@
...
+func getConnectionIDFromMessage(message *ws.ProtoMsg) string {
+	connectionID, _ := message.Header.Properties[propertyConnectionID].(string)
+	return connectionID
+}
+
+func (d *MenderShellDaemon) routeMessagePortForwardNew(message *ws.ProtoMsg) error {
+	connectionID := getConnectionIDFromMessage(message)
+	protocol := message.Header.Properties[propertyProtocol].(string)
+	host := message.Header.Properties[propertyRemoteHost].(string)
+	portNumber := message.Header.Properties[propertyRemotePort].(uint64)
```
kacf, line 166: All these three rely on external network input and can crash the daemon if the type doesn't match.

kacf (commented):
Nice work! It really is quite small; a lot of value per line of code!

This needs integration tests though, and depends on mender-cli changes. Therefore I think this should be targeted towards a feature branch, so it doesn't get merged before we have all of them ready.

config/config.go
tranchitella: This is what we decided to implement for all the other features: opt-out.
See: https://github.com/mendersoftware/mender-connect/blob/master/config/config.go#L39

app/daemon_portforward.go
tranchitella: This was temporary, mainly because go-lib-micro wasn't ready.
They moved there, now. :)

app/daemon_portforward.go
tranchitella: You are right, fixed!

kacf (approved):
Although a lot of things moved, I assume this is mostly adjustments to follow latest code changes and review comments, so I didn't go into quite as much detail as last time. Looks good to me!

session/portforward.go
```diff
@@ -0,0 +1,276 @@
...
+		err = portForwardHandlerForward(msg, w)
+	default:
+		err = errPortForwardUnkonwnMessageType
+	}
+	if err != nil {
+		log.Errorf("portForwardHandler(%+v)", err)
+
+		response := &ws.ProtoMsg{
+			Header: ws.ProtoHdr{
+				Proto:     ws.ProtoTypeControl,
```
alfrunes, line 178: The handler cannot speak on behalf of another protocol.
```suggestion
				Proto:     ws.ProtoTypePortForward,
```

session/portforward.go
```diff
@@ -0,0 +1,276 @@
...
+				MsgType:   ws.MessageTypeError,
+				SessionID: msg.Header.SessionID,
+			},
+			Body: []byte(err.Error()),
```
alfrunes, line 182: I'd recommend adding an error schema to keep the API more flexible (and easier to add documentation).

session/portforward.go
```diff
@@ -0,0 +1,276 @@
...
+		return errPortForwardInvalidMessage
+	}
+
+	portForwarder := &MenderPortForwarder{
+		ConnectionID:   connectionID,
+		SessionID:      message.Header.SessionID,
+		ResponseWriter: w,
+	}
+
+	key := message.Header.SessionID + "/" + connectionID
```
alfrunes, line 211: Not necessary to use the SessionID as it will be the same for one instance of this handler.

session/portforward.go
```diff
@@ -0,0 +1,276 @@
...
+	f.ctxCancel = cancelFunc
+
+	go f.Read()
+
+	return nil
+}
+
+func (f *MenderPortForwarder) Close() error {
+	log.Debugf("port-forward[%s/%s] close", f.SessionID, f.ConnectionID)
+	f.ctxCancel()
```
alfrunes, line 79: maybe I'm missing it, but I cannot see that the entries in the global `portForwarders` belonging to this session instance gets cleaned up.

tranchitella: there are `delete(portForwarders, key)` on connection closing (or errors). Do you mean something else?

alfrunes: `Close` is called whenever the session is closed - whether it's due to a fatal error or regular close. So if the connection times out, mender-connect is keeping the map entry in memory until the application is closed.
The easiest way (I can think of) is to let the garbage collector do the work by putting the global variable inside the `MenderPortForwarder` struct.

tranchitella: The Close method is now responsible for removing the port forwarder from the global variable, does it make sense?

alfrunes: The only issue is that you can't send messages in the `Close` handler, as it's called after the Session close condition is met. It should only deallocate resources (sorry for not making it clear earlier).

lluiscampos:
The last commit deserves a `Changelog` tag

> (...) Therefore I think this should be targeted towards a feature branch, so it doesn't get merged before we have all of them ready.

I also think we should go via feature branch this time, so that we can possibly release without it if we run out of time.

session/portforward.go
tranchitella: I believe you are confusing my MenderPortForwarder (which is an internal struct to hold a single port-forwarding session) with the Session handler. Can it be?

alfrunes: Yeah, sorry... My bad, I thought that `MenderPortForwarder` was implementing the `SessionHandler` interface, but I see now that you are using the `HandlerFunc` wrapper, so `Close` is indeed not the `Close` from the `SessionHandler` interface. I'll try to be more thorough next time.

tranchitella:

session/portforward.go
```diff
@@ -0,0 +1,335 @@
...
+	log.Debugf("port-forward[%s/%s] write %d bytes", f.SessionID, f.ConnectionID, len(body))
+	_, err := f.conn.Write(body)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func PortForward() Constructor {
+	f := HandlerFunc(portForwardHandler)
```
alfrunes, line 173: `HandlerFunc` was intended for stateless handlers that doesn't consume external resources. The problem with using it here is that it does not free the local socket & port on the device if the point-to-point connection closes unexpectedly. The `SessionHandler.Close` is called whenever the health check for the `Session` fails, and should be responsible for cleaning up allocated resources. IOW; convert `MenderPortForwarder` to a `SessionHandler` by adding a `Close() error` member and make `portForwardHandler` a member function with the name `ServeProtoMsg(*ws.ProtoMsg, ResponseWriter)`.

tranchitella: @alfrunes something like this? https://github.com/mendersoftware/mender-connect/pull/15/commits/0fe7762f407d259b112527f1c57d6824c476a669

session/portforward.go
```diff
@@ -169,22 +168,36 @@ func (f *MenderPortForwarder) Write(body []byte) error {
...
+	return func() SessionHandler {
+		return &PortForwardHandler{
+			portForwarders: make(map[string]*MenderPortForwarder),
+		}
+	}
+}
+
+func (h *PortForwardHandler) Close() error {
+	for _, f := range h.portForwarders {
+		f.Close(true)
```
alfrunes, line 182: This function is called after a session close condition and no messages should be sent nor received.
```suggestion
		f.Close(false)
```

session/portforward.go
```diff
@@ -234,15 +247,16 @@ func portForwardHandlerNew(message *ws.ProtoMsg, w ResponseWriter) error {
 		ConnectionID:   connectionID,
 		ResponseWriter: w,
 		mutexAck:       &sync.Mutex{},
+		portForwarders: h.portForwarders,
 	}
 
 	key := message.Header.SessionID + "/" + connectionID
-	portForwarders[key] = portForwarder
+	h.portForwarders[key] = portForwarder
```
alfrunes: `SessionID` will be the same for all entries in this map now, so we might as well index by the connectionID.
```suggestion
	h.portForwarders[connectionID] = portForwarder
```

alfrunes (changes requested):
I only have one small request then I'll stop whining :)

session/portforward.go
tranchitella: right, sorry, I did not remember it

session/portforward.go
tranchitella: Right! https://github.com/mendersoftware/mender-connect/pull/15/commits/b666f62245939059bcf0c89369adea52fa3665fa

tranchitella:
> I only have one small request then I'll stop whining :)

No need to stop, it wasn't obvious to me how to use the Session Handler and your reviews are extremely useful.

alfrunes (approved):

PR #16, opened 2021-02-07
Add remote triggers for Mender client's check-update and send-inventory

tranchitella:
This PR pairs with https://github.com/mendersoftware/deviceconnect/pull/38


tranchitella:
Failure `tests/test_fault_tolerance.py::TestFaultTolerance::test_rootfs_conf_missing_from_new_update` is spurious.

app/daemon_menderclient.go
```diff
@@ -0,0 +1,69 @@
...
+	switch message.Header.MsgType {
+	case menderclient.MessageTypeMenderClientCheckUpdate:
+		killSignal = "-USR1"
+	case menderclient.MessageTypeMenderClientSendInventory:
+		killSignal = "-USR2"
+	}
+	if killSignal != "" {
+		return sendSignalToProcess(
+			exec.Command("kill", killSignal),
+			exec.Command("systemctl", "show", "-p", "MainPID", "mender-client"))
```
kacf, line 38: Please use `mender check-update` and `mender send-inventory` instead. This has been a notoriously buggy part in the client, so it would be better to reuse that. I'm not even completely sure if we will keep USR1 and USR2 signals after the ticket in the next paragraph is implemented.

Note [MEN-4194](https://tracker.mender.io/browse/MEN-4194), which is going to provide a better interface over DBus. Due to customer requirements this is likely to be scheduled rather soon. I'm not sure if it influences the design of the Connect protocol, but we should consider that the request may return information as well in the future. How is that best handled? Just a new request code later on, or should we think about how this one can be extended later?

kacf (changes requested):

lluiscampos (commented):
Nice :heart: 

A part from what Kristian has said (I support that), one more note:

I think this functionality deserves an integration test, as involves the client, mender-connect and several micro-services. Any plans for that?

app/daemon_menderclient.go
tranchitella: Ok, I'll rework to call directly the client. Regarding your question, I know MEN-494 provides a DBus interface and we'll switch to it as soon as it is ready. The reason why we need something like this is the fact we'll use it in the first iteration of the single device configuration epic. This will allow us to avoid waiting for the next deployments/next call according to the polling interval to deploy the configuration.

tranchitella:
> Nice ❤️
> A part from what Kristian has said (I support that), one more note:
> I think this functionality deserves an integration test, as involves the client, mender-connect and several micro-services. Any plans for that?

Yes, as I mentioned above, this is going to be part of the single device configuration deployment.

app/daemon_menderclient.go
kacf: Ok, I guess we will consider adding a new request code as an extension later then.

lluiscampos:
> > [...] Any plans for that?
> 
> Yes, as I mentioned above, this is going to be part of the single device configuration deployment.

Oh, so this is part of the epic! I miss-understood then. Sorry.

app/daemon_menderclient.go
tranchitella: I'm not sure I get what you mean with "request code": is it the message type in the ws.ProtoMsg struct? In this PR, mender-connect doesn't reply back to device-connect, thus no output/response is provided. We can add it using the same message type, though, either as Body or using the properties map (key/value pairs).

tranchitella: I reimplemented it to call the mender client

lluiscampos (approved):
Looks good to me now!

tranchitella:
@kacf new round of review?

app/daemon_menderclient.go
kacf: > In this PR, mender-connect doesn't reply back to device-connect, thus no output/response is provided.

Ah, but this it should probably do. It should reply with some sort of OK, so that we can tell from the connecting side whether the call worked. This is important for compatibility in both directions:

* If a new mender-cli calls mender-connect from Mender 2.6.0, it will receive an error for this message type. If it sends it to mender-connect from Mender 2.7.0 however, it will receive nothing. So did it work, or did the connection break?
* Similarly, if mender-cli calls mender-connect from Mender 2.8.0, it will receive a response with a body, but if it's 2.7.0, it receives nothing and can't tell whether it worked or not.

Does it make sense?

> I reimplemented it to call the mender client

Perfect!

alfrunes (approved):

kacf (approved):
Looks good!

app/daemon_menderclient.go
kacf: Do you have any thoughts on this comment? It's not directly related to this PR, but I think it's quite important for future compatibility that we use a request/response pattern, at least for the initial message.

tranchitella: mender-cli is not involved in this call; it's deviceconnect which translates an API call to a message to the device:
https://github.com/mendersoftware/deviceconnect/blob/master/api/http/internal.go#L109

We can send back the OK message, but in the current implementation of deviceconnect nothing is waiting for it. The message doesn't even set the session ID, which means the message cannot be retrieved by this particular API handler. We can extend the implementation if needed, but from the back-end point of view, this is more a "fire-and-forget" thing: if it works, it's better, but if it doesn't it is not an issue at all.

Does it make sense?

kacf: I understand your point, but I think that all introductory messages in the ProtoMsg protocol should have an ACK message, because ProtoMsg is supposed to be generic. This is the only way that we can probe for capabilities.

> We can send back the OK message, but in the current implementation of deviceconnect nothing is waiting for it.
> ...
> this is more a "fire-and-forget" thing

This is not completely true. If you send this message type to mender-connect from Mender 2.6.0, then you will get a response, an error. So if you use mender-cli to send `ProtoTypeMenderClient` message, how long should it wait for an error? 1 second? 5 seconds? Either it will receive an error, or it will receive silence, so it has no choice but to wait for a long time to confirm.

Think of it this way: If you send a POST request to an HTTP endpoint, it would be unacceptable that it returns not a single byte in return. This feels like an implementation bug, and you are no wiser as to what happened to your request. It should return 204, or some error.

The fact that nothing is waiting for the OK message *today*, is fine. But we should still send it, because the question is what will be waiting for it tomorrow. We have to assume that old and new versions will be mixed.

tranchitella: @kacf I changed my mind, you are correct, let's introduce the confirmation

tranchitella: https://github.com/mendersoftware/mender-connect/pull/16/commits/bbecbddbdd0ddc71f41c59120e8f73cff650ef91

kacf: :+1:

PR #17, opened 2021-02-19
MEN-4322: File transfer protocol

alfrunes:


session/session.go
```diff
@@ -11,447 +11,199 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+
 package session
```
alfrunes, line 15: Moved shell session to `shell.go` and recreated `session.go` with the new `Session` definition and handler interface, hence the enormous diff.

merlin-northern:
@alfrunes it says WIP in the desc, so is that ready for review?

alfrunes:
> @alfrunes it says WIP in the desc, so is that ready for review?

Now it doesn't :see_no_evil:

alfrunes:
start pipeline

session/session.go
```diff
@@ -11,447 +11,202 @@
...
+			sess.Error(&ws.ProtoMsg{}, true, "session: timed out")
 			return
-		case <-s.pong:
-			s.healthcheckTimeout = time.Now().Add(healthcheckInterval + healthcheckTimeout)
-		case <-time.After(time.Until(s.healthcheckTimeout)):
-			if s.healthcheckTimeout.Before(time.Now()) {
-				log.Errorf("session %s, health check failed, connection with the client lost", s.id)
-				s.expiresAt = time.Now()
+
+		case <-time.After((sess.Config.IdleTimeout * 4) / 5):
```
merlin-northern, line 209: some comment why *0.8 ?

session/filetransfer.go
```diff
@@ -419,6 +421,30 @@ func (h *FileTransferHandler) FileUploadHandler(
 		done bool
 		open bool
 	)
+	writeChunk := func(msg *ws.ProtoMsg) error {
```
merlin-northern, line 424: is there really a reason for it to not be a regular function? we can have units for it even :)

merlin-northern (commented):
one general remark: are we checking the free disc space? high tech: we could check when we flush every other chunk , or check at the beginning.

config/config.go
```diff
@@ -35,6 +35,13 @@ const httpsSchema = "https"
 type TerminalConfig struct {
 	Width  uint16
 	Height uint16
+	// FIXME (unused)
+	Disable bool
```
merlin-northern, line 39: at least two more words, why it is here then?

session/filetransfer.go
```diff
@@ -0,0 +1,496 @@
...
+	for ackOffset <= chunker.Offset {
+		select {
+		case msg, open = <-h.msgChan:
+			if !open {
+				return nil
+			}
+			if msg.Header.MsgType != wsft.MessageTypeACK {
+				// This should NEVER happen
+				err = errors.New("internal error")
+				return err
```
merlin-northern, line 274:
```suggestion
				return errors.New("internal error")
```

merlin-northern: how about a log with Error here as well?

session/filetransfer.go
```diff
@@ -0,0 +1,496 @@
...
+			}
+			if off, ok := msg.Header.Properties["offset"].(int64); ok {
+				ackOffset = off
+			} else {
+				ackOffset += BufSize
+			}
+		case <-time.After(time.Second * 10):
+			err = errors.New("timed out waiting for ACK message")
+			return err
```
merlin-northern, line 255:
```suggestion
			return errors.New("timed out waiting for ACK message")
```
and perhaps log with Error?

session/filetransfer.go
```diff
@@ -0,0 +1,496 @@
...
+		case msg, open = <-h.msgChan:
+			if !open {
+				return nil
+			}
+			if msg.Header.MsgType != wsft.MessageTypeACK {
+				// This should NEVER happen
+				err = errors.New("internal error")
+				return err
+			}
+			if off, ok := msg.Header.Properties["offset"].(int64); ok {
```
merlin-northern, line 276: a helper
```go
func getOffset(msg *ws.MsgProto) int64,bool {
}
```
maybe?

session/filetransfer.go
```diff
@@ -0,0 +1,496 @@
...
+		rsp := &ws.ProtoMsg{Header: msg.Header}
+		rsp.Header.MsgType = wsft.MessageTypeACK
+		w.WriteProtoMsg(rsp)
+	}
+	filename := fd.Name()
+	errClose := fd.Close()
+	if errClose != nil {
+		log.Warnf("error closing file: %s", errClose.Error())
+	}
+	err = os.Rename(filename, *params.Path)
```
merlin-northern, line 494: error on rename seems more fit for a log line with error than the error on closing, which happens rarely.

session/session.go
```diff
@@ -11,447 +11,237 @@
...
+func (sess *Session) ListenAndServe() {
+	defer sess.handlePanic()
+	var (
+		msg      *ws.ProtoMsg
+		open     bool
+		deadline time.Time = time.Now().Add(sess.Config.IdleTimeout)
+	)
+	select {
+	case <-sess.done:
+		panic("session already finished")
```
merlin-northern, line 200: are we sure we want to panic here?

session/filetransfer.go
```diff
@@ -0,0 +1,496 @@
...
+				if err == io.EOF {
+					done = true
+				} else if err != nil {
+					return err
+				}
+			default:
+				break Loop
+			}
+		}
+		if i == 0 {
```
merlin-northern, line 466: I do not understand how and when we can enter this if?

session/filetransfer.go
```diff
@@ -0,0 +1,549 @@
...
+	} else if err = params.Validate(); err != nil {
+		return errors.Wrap(err, "invalid request parameters")
+	}
+
+	go h.FileUploadHandler(msg, params, w) //nolint:errcheck
+	return nil
+}
+
+func prepareTempFile(params FileInfo) (*os.File, error) {
+	fd, err := ioutil.TempFile(path.Dir(*params.Path), path.Base(*params.Path))
```
merlin-northern, line 372: two questions here:
1. is the file removed in all scenarios before we reach the final move to the target path?
2. this is not safe, since TempFile creation and the Chown below do not make an atomic operation; suggested change:
* set umask to `0077`
* create a temporary directory
* create all the temporary files in that directory
* once done make the final move, checking if the destination is not something suspicious
* remove the temporary directory recursively
* restore the umask
* set the modes and permissions on the destination file

alfrunes:

app/daemon.go
```diff
@@ -84,11 +84,22 @@ type MenderShellDaemon struct {
...
+	routes := make(session.ProtoRoutes)
+	if config.FileTransfer.Enable {
+		routes[ws.ProtoTypeFileTransfer] = session.FileTransfer()
+	}
+	router := session.NewRouter(
+		routes, session.Config{
+			IdleTimeout: connectionmanager.DefaultPingWait,
+		},
+	)
+
```
merlin-northern, line 102: this needs explaining in the comments.

session/filetransfer.go
alfrunes: 1. Good catch, this is indeed not the case if this function returns an error.
2. I'll try to avoid using `ioutil.Tempfile`, but I don't think it's a good idea to create the file outside the target directory (since we don't know if the target is on a different filesystem; nor if we have permission to create the file there at all).

alfrunes: The solution I went for is creating a new temporary file in the target directory with `0200` permissions. Then, before I commit the file to the target path, I change the permissions and owner (in that order) to the ones provided in the initial request.

alfrunes:

session/session.go
```diff
@@ -11,447 +11,243 @@
...
+			constructor, ok := sess.Routes[msg.Header.Proto]
+			if !ok {
+				sess.Error(msg, false, fmt.Sprintf(
+					"no handler registered for protocol: 0x%04X",
+					msg.Header.Proto,
+				))
+				continue
+			}
+			handler = constructor()
+			defer handler.Close()
```
merlin-northern, line 248: just for my peace of mind, can you confirm that the Close() method will have been called on exit from this function for every handler?

session/session.go
```diff
@@ -11,447 +11,243 @@
...
+			fmt.Fprintf(&stacktrace, "\n%s:%d.%s",
+				file, line, funcname(fn.Name()),
+			)
+		}
+		log.WithField("trace", stacktrace.String()).
+			Errorf("[panic] %s", r)
+		sess.Error(&ws.ProtoMsg{}, true, "internal error")
+	}
+	close(sess.done)
+}
```
merlin-northern, line 189: this is slightly too much for me. why are we doing that?

session/session.go
```diff
@@ -11,447 +11,243 @@
...
+
+func (sess *Session) ListenAndServe() {
+	const pongWait = time.Second * 5
```
merlin-northern, line 192: maybe only me but I feel that this is buried here, just confirm that this is a good place for this const and that `5s` is not used anywhere else.

session/session.go
```diff
@@ -11,447 +11,243 @@
...
-	}
-	s.pseudoTTY.Close()
+		if msg.Header.Proto == ws.ProtoTypeControl {
+			if sess.HandleControl(msg) {
+				return
+			}
+			continue
+		}
 
-	err = procps.TerminateAndWait(s.shellPid, s.command, 2*time.Second)
```
merlin-northern, line 450: just for my peace of mind (sorry, again) where did that go?

session/filetransfer.go
alfrunes: Sorry, this is not really explained in the code. The above loop will consume all inbound messages that are readily available without blocking on I/O. The `i==0` is only the case if no messages are pushed on the channel, and we need to block for messages.

app/daemon.go
alfrunes: Yup, I forgot to put comments explaining this, but I did a modest attempt in the commit messages.

session/session.go
alfrunes: This was mostly taken from the `accesslog` middleware. I don't we should terminate on panic in handler functions - simply closing and freeing the session should be enough. The stacktrace logger field is simply for debugging purposes.

session/session.go
alfrunes: `session/shell.go`

session/session.go
alfrunes: I'll bring out the constants to the global scope (I agree, it's pretty messy to keep it local to the functions). This constant in particular is how long we wait for a pong after sending a ping (after which the session times out).

session/session.go
alfrunes: The Close function is pushed on the defer stack as soon as the handler is created. So under normal shutdown it should always call the close function. However, I did not consider exceptional cases such as terminating signals etc. Do you think it's necessary to add a signal handler that ensures all resources are deallocated?

app/daemon.go
```diff
@@ -84,11 +84,22 @@ type MenderShellDaemon struct {
...
 	shellsSpawned           uint
 	debug                   bool
+	router                  session.Router
 }
 
 func NewDaemon(config *configuration.MenderShellConfig) *MenderShellDaemon {
 	ctx, ctxCancel := context.WithCancel(context.Background())
 
+	routes := make(session.ProtoRoutes)
+	if config.FileTransfer.Enable {
```
tranchitella, line 94: I'm not 100% sure, but I believe the discussion led us to enable file transfer by default, and provide an option to disable it. This way, it will work out-of-the-box without any special configuration (e.g. onboarding).

session/filetransfer.go
```diff
@@ -0,0 +1,580 @@
...
+			h.Error(msg, w, err)
+		}
+	}()
+	if err := msgpack.Unmarshal(msg.Body, params); err != nil {
+		err = errors.Wrap(err, "malformed request parameters")
+		return err
+	} else if err = params.Validate(); err != nil {
+		err = errors.Wrap(err, "invalid request parameters")
+		return err
+	}
```
tranchitella, line 235: I'm not 100% sure, but I believe you are not writing the "err" variable returned to the function inside the "if" for error handling of msgpack.Unmarshal, because you are initializing a new variable (:=).

session/filetransfer.go
```diff
@@ -0,0 +1,580 @@
...
+		}
+		return msg, nil
+	}
+
+	buf := make([]byte, FileTransferBufSize)
+	for {
+		windowBytes := ackOffset - chunker.Offset +
+			ACKSlidingWindowRecv*FileTransferBufSize
+		if windowBytes > 0 {
+			N, err := io.CopyBuffer(chunker, io.LimitReader(fd, windowBytes), buf)
```
tranchitella, line 314: Same here, you are defining a new var, aren't you?

tranchitella (approved):

app/daemon.go
alfrunes: Changed to `!config.FileTransfer.Diable`

session/filetransfer.go
alfrunes: You're right. However, I think that the `return err` assigns the new variable to the old one. I changed it just to be sure.

alfrunes:

alfrunes:

session/filetransfer.go
alfrunes: I refactored this bit to make it a bit clearer.

alfrunes:
Started a [new pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/267114397) together with the PRs:
 * mendersoftware/integration#1251
 * mendersoftware/deviceconnect#51

session/filetransfer.go
```diff
@@ -0,0 +1,587 @@
...
+func (h *FileTransferHandler) writeFile(w ResponseWriter, dst *os.File) (int64, error) {
+	var (
+		done   bool
+		open   bool
+		err    error
+		i      int
+		offset int64
+		msg    *ws.ProtoMsg
+	)
+	// Convenience clojure for decoding file chunk and writing to destination file.
```
merlin-northern, line 503: is "clojure" an intentional reference to Lisp?

alfrunes: Oops, that should say "function closure" :see_no_evil:

tranchitella (approved):
I reviewed the last three commits.

PR #24, opened 2021-04-07
stat(2) file path and verify conditions before starting upload

alfrunes:
This is not the complete implementation that we talked about during the backend team meeting. I realized that adding the filename parameter will require updates to all dependencies as well, as this will introduce a new model in go-lib-micro (right now we're reusing the FileInfo model for the upload).

### DIFF
--- a/session/filetransfer.go
+++ b/session/filetransfer.go
@@ -393,6 +393,18 @@ func (h *FileTransferHandler) InitFileUpload(msg *ws.ProtoMsg, w ResponseWriter)
 		return filetransfer.ErrTxBytesLimitExhausted
 	}
 
+	if info, errStat := os.Lstat(*params.Path); errStat != nil {
+		if !os.IsNotExist(errStat) {
+			err = errors.Wrap(errStat, "error checking file location")
+			return err
+		}
+	} else if !info.Mode().IsRegular() {
+		err = errors.New(
+			"conflicting file path: cannot overwrite irregular file",
+		)
+		return err
+	}
+
 	select {
 	case h.mutex <- struct{}{}:
 		go h.FileUploadHandler(msg, params, w) //nolint:errcheck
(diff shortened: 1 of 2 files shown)


alfrunes:
If we want the source path as an additional parameter I've added a PR updating the model here: mendersoftware/go-lib-micro#126

session/filetransfer.go
```diff
@@ -393,6 +393,18 @@ func (h *FileTransferHandler) InitFileUpload(msg *ws.ProtoMsg, w ResponseWriter)
...
+			err = errors.Wrap(errStat, "error checking file location")
+			return err
+		}
+	} else if !info.Mode().IsRegular() {
+		err = errors.New(
+			"conflicting file path: cannot overwrite irregular file",
+		)
+		return err
+	}
+
```
merlin-northern, line 407: we do have [utils/fs.go](https://github.com/mendersoftware/mender-connect/blob/master/utils/fs.go#L41) maybe lets extend that?

alfrunes: We need the error returned from stat as well in case the file does not exist.

session/filetransfer_test.go
```diff
@@ -100,33 +100,41 @@ func TestFileTransferUpload(t *testing.T) {
...
-				Counters: config.RateLimits{
-					MaxBytesTxPerMinute: 1,
-					MaxBytesRxPerMinute: 1,
-				},
-			},
-
-			Error: errors.New("no file transfer in progress"), //for some reason Upload currently returns this error
+			Error: errors.New("conflicting file path: cannot overwrite irregular file"),
 		},
+		// The following test case does not seem to be implemented yet.
```
merlin-northern, line 110: I must followup on this -- the limits are in place, but they need a new unit, they will be covered with full integration test, so lets just remove it for now.

merlin-northern (approved):
lgtm I know you are going to make good use of the comments :)

alfrunes:

tranchitella (approved):

PR #31, opened 2021-04-20
File transfer limits: use of the proper flag for nosuid.

merlin-northern:
ChangeLog:none


alfrunes:
From internal meeting:
* This will break the bidirectional protocol API.
  * Will keep the permission bit masking currently in `2.7` and ignore other file mode bits.

merlin-northern:
> From internal meeting:
> 
> * This will break the bidirectional protocol API.

it will not break it, the only thing would be: stat endpoint would return the different modes than the ones specified with mode field in upload.

this was just a fix for the limits and it was all that was needed to have mode=4755 fully functional.

PR #42, opened 2021-07-08
[MEN-4864] Try refreshing JWT token if server returns 401 (unauthorized)

alfrunes:
This commit merges the dbusEventLoop and eventLoop into a single
eventLoop routine, since separating the reconnection logic can lead to
unintended short-circuit loops if the device becomes decommissioned.

changelog: title

Signed-off-by: Alf-Rune Siqveland <alf.rune@northern.tech>

PR #37, opened 2021-06-23
MEN-4694 Break reconnection loop on auth lost

merlin-northern:
ChangeLog:none
Signed-off-by: Peter Grzybowski <peter@northern.tech>

PR #44, opened 2021-07-09
Revert "Merge pull request #42 from alfrunes/MEN-4864"

alfrunes:
This reverts commit 8a76f5930ecff4137d525ed3b0382853ccebe381, reversing
changes made to 48044ce5f0e6921a6ea72668f280590195b3108f.

@lluiscampos False alarm, after trying to debug for several hours I realized that I have been using mender-connect 1.1.0 without knowing :man_facepalming: 
In any case, I am unable to reproduce the issue after #37 was merged.

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -55,6 +55,7 @@ type MenderShellDaemon struct {
 	writeMutex              *sync.Mutex
 	eventChan               chan MenderShellDaemonEvent
 	connectionEstChan       chan MenderShellDaemonEvent
+	reconnectChan           chan MenderShellDaemonEvent
 	stop                    bool
 	authorized              bool
 	printStatus             bool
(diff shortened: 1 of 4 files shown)


lluiscampos:
@alfrunes Missing the signature in the commit and (I think!) the cancel changelog tag.

@kacf It might be that it works when reverting a commit, but not when reverting a PR using GitHub UI.

alfrunes:
> @alfrunes Missing the signature in the commit and (I think!) the cancel changelog tag.
> 
> @kacf It might be that it works when reverting a commit, but not when reverting a PR using GitHub UI.

@lluiscampos I used `git` cli to revert the merge commit:
`git revert -m 1 8a76f5930ecff4137d525ed3b0382853ccebe381`

lluiscampos:
> > @alfrunes Missing the signature in the commit and (I think!) the cancel changelog tag.
> > @kacf It might be that it works when reverting a commit, but not when reverting a PR using GitHub UI.
> 
> @lluiscampos I used `git` cli to revert the merge commit:
> `git revert -m 1 8a76f5930ecff4137d525ed3b0382853ccebe381`

Okey, I was wrong about the process. But the point is that the syntax is not the same as when reverting non-merge commit, so I guess the tool does not support this case.

kacf:
> @lluiscampos I used `git` cli to revert the merge commit:
> `git revert -m 1 8a76f5930ecff4137d525ed3b0382853ccebe381`

No need to fix it now, but it's better to revert the original commit I think. This is supported by the tool, and gives better commit messages IMHO, since it's the original message which gets included.

PR #46, opened 2021-11-11
MEN-5203: Remove redundant server and client configuration

lluiscampos:
This configuration is now handled by the mender client
    
    Changelog: Deprecate configuration fields `ServerURL` , `Servers`,
    `ServerCertificate`, `ClientProtocol`, `HTTPSClient` and `SkipVerify`.
    The server configuration is obtained via D-Bus API, for which from
    Mender 3.2 onwards is a local proxy which uses `mender.conf` client
    settings to securely authenticate with the backend server, making the
    client settings in `mender-connect.conf` redundant.


lluiscampos:
@mender-test-bot start pipeline :robot:

lluiscampos:
New commit:

```
    MEN-5203: Deprecate advance client configuration
    
    This configuration is now handled by the mender client
    
    Cancel-changelog: 01bce7ccd942ff6471dded7581c9bc2cf14787c7
    
    Changelog: Deprecate configuration fields `ServerURL` , `Servers`,
    `ServerCertificate`, `ClientProtocol`, `HTTPSClient` and `SkipVerify`.
    The server configuration is obtained via D-Bus API, for which from
    Mender 3.2 onwards is a local proxy which uses `mender.conf` client
    settings to securely authenticate with the backend server, making the
    client settings in `mender-connect.conf` redundant.
```

lluiscampos:
@mender-test-bot start pipeline :duck:

config/config.go
```diff
@@ -116,14 +112,6 @@ type Limits struct {
...
 // MenderShellConfigFromFile holds the configuration settings read from the config file
 type MenderShellConfigFromFile struct {
-	// ClientProtocol "https"
-	ClientProtocol string
-	// HTTPS client parameters
-	HTTPSClient https.Client `json:"HttpsClient"`
-	// Skip CA certificate validation
-	SkipVerify bool
-	// Path to server SSL certificate
-	ServerCertificate string
```
merlin-northern, line 128: this is a separate ticket, which has nothing to do with MEN-5203. it is removing https support from mender-connect. if we have decided that mender-connect now uses only http it is fine, but otherwise, the ServerCertificate should be provided over D-Bus as ServerURL is. In that case we should keep SkipVerify (or also pass it via D-Bus).

kacf: It wasn't originally mentioned in MEN-5203, but @lluiscampos and me discussed this on Slack. The reason we want to remove it is that the D-Bus API will never return an https address anymore. It will always be an http link to the local proxy. Hence there is no need for anything certificate related in mender-connect.

kacf (approved):
:+1: More than 200 lines removed!

config/config.go
lluiscampos: @merlin-northern I will extend the acceptance criteria on the ticket to cover this. I also made it on a separate commit to leave the discussion open.

connection/connection.go
```diff
@@ -17,8 +17,6 @@ package connection
 import (
 	"crypto/tls"
 	"crypto/x509"
```
lluiscampos, line 19: Note to self. We probably can get the rid of these dependencies too, `loadServerTrust` should not be needed neither.

config/config.go
lluiscampos: Ticket updated, commits squashed now.

lluiscampos:
@mender-test-bot start pipeline :see_no_evil:

config/config.go
```diff
@@ -355,11 +323,47 @@ func loadConfigFile(configFile string, config *MenderShellConfig, filesLoadedCou
...
+		log.Errorf("Error loading configuration from file: %s (%s)", configFile, err.Error())
+		return err
+	}
+
 	*filesLoadedCount++
 	log.Info("Loaded configuration file: ", configFile)
 	return nil
 }
 
+func checkforDeprecatedFields(configFile string) error {
```
merlin-northern, line 336: this one does not have units -- hence the coverage decrease I guess.

config/config_test.go
```diff
@@ -15,26 +15,19 @@
...
 	"os"
 	"path"
 	"reflect"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	log "github.com/sirupsen/logrus"
 
-	"github.com/mendersoftware/mender-connect/client/https"
+	"github.com/stretchr/testify/assert"
```
merlin-northern, line 27: if I read this diff correctly, there is an extra empty line here that is not needed.

merlin-northern (approved):

config/config.go
lluiscampos: Most of it [is tested](https://github.com/mendersoftware/mender-connect/pull/46/files#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R372), and so it is reflected in the [coverage report](https://coveralls.io/jobs/89987608/source_files/14424565085#L336).

config/config_test.go
lluiscampos: It seems like. But the space is pre-existing and to have the `testify` packages separated is very common in test files. I'll leave it.

lluiscampos:
Last force-push only brings a rebase from master. Not re-running the pipeline.

PR #47, opened 2021-11-17
pipeline: deactivate child pipelines when running on schedule

lluiscampos:
For weekly scheduled builds, we don't want to trigger child pipelines
that just flood our CI infra. Specially for mender-dist-packages, where
each pipeline builds/tests/publishes all packages.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -44,6 +44,8 @@ generate-qa-trigger:
     - if: '$CI_COMMIT_BRANCH =~ /^(master|[0-9]+\.[0-9]+\.x)$/'
     - if: '$CI_PIPELINE_SOURCE == "pipeline"'
       when: never
+    - if: '$CI_PIPELINE_SOURCE == "schedule"'
+      when: never
   before_script:
     - apk add --no-cache git
     - pip3 install pyyaml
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mzedel Before I port this to `mender`, `mender-configure-module` and `monitor-client`... should it also be added to `trigger:mender-qa` job?

oleorhagen (approved):

.gitlab-ci.yml
```diff
@@ -70,6 +72,8 @@ trigger:mender-dist-packages:
   stage: trigger
   rules:
     - if: '$CI_COMMIT_BRANCH == "master"'
+    - if: $CI_PIPELINE_SOURCE == "schedule"
```
mzedel, line 75:
```suggestion
    - if: '$CI_PIPELINE_SOURCE == "schedule"'
```
please align these quotes to be consistent with the rest of the file

mzedel (changes requested):

mzedel:
> @mzedel Before I port this to `mender`, `mender-configure-module` and `monitor-client`... should it also be added to `trigger:mender-qa` job?

as far as I understand the trigger would include the current repo state (so... `master`) and run the qa pipeline with it - for schedules that should be covered by the scheduled run of the `mender-qa` pipeline I think... if we run scheduled pipelines for different combinations it might be better to include it, but in the general `master` case I think not including it is fine.

.gitlab-ci.yml
lluiscampos: Right, thanks for caring about these details :)

lluiscampos:
> @mzedel Before I port this to `mender`, `mender-configure-module` and `monitor-client`... [...]

Launched now the same fix in the other repos:
* https://github.com/mendersoftware/mender/pull/889
* https://github.com/mendersoftware/mender-configure-module/pull/37
* https://github.com/mendersoftware/monitor-client/pull/96

mzedel (approved):

PR #48, opened 2021-11-22
QA-235: pipeline: Switch to static checks by golangci-lint

lluiscampos:
And fix the code accordingly.


lluiscampos:
@mender-test-bot start pipeline :shark:

connection/connection.go
```diff
@@ -61,10 +54,11 @@ func NewConnection(u url.URL,
 
 	headers := http.Header{}
 	headers.Set("Authorization", "Bearer "+token)
-	ws, _, err := dialer.Dial(u.String(), headers)
+	ws, rsp, err := dialer.Dial(u.String(), headers)
 	if err != nil {
 		return nil, err
 	}
+	defer rsp.Body.Close()
```
merlin-northern, line 61: I am not sure about that one. was this required by the static check? or do you think it is required?

lluiscampos: I was required by the static check, same as we have done [for `mender-cli`](https://github.com/mendersoftware/mender-cli/pull/139/files#diff-3d829077a7d1516e689ac2ac153f9caaf9e4a0004c4c9d059c4800ec279af26dR104).

What are you unsure about? `dialer.Dial` does an HTTP request and [returns the response](https://pkg.go.dev/github.com/gorilla/websocket#Dialer.Dial) before the upgrade to websockets. This response's body [has to be closed](https://pkg.go.dev/net/http#ReadResponse).

limits/filetransfer/limits.go
```diff
@@ -373,8 +377,7 @@ func updateCounters() {
 	tick := time.NewTicker(time.Duration(countersUpdateSleepTimeS) * time.Second)
 	defer tick.Stop()
 	for counterUpdateRunning {
-		select {
-		case <-tick.C:
+		for range tick.C {
```
merlin-northern, line 380: was this also required by the static checker? as far as I know it should work, its just this part of code was tested manually for such a long time, including some graphs of the tx/rx from debug info :) you know. if required let's have it.

session/portforward.go
```diff
@@ -35,7 +35,9 @@ const (
 )
 
 var (
-	errPortForwardInvalidMessage     = errors.New("invalid port-forward message: missing connection_id, remort_port or protocol")
+	errPortForwardInvalidMessage = errors.New(
+		"invalid port-forward message: missing connection_id, remort_port or protocol",
```
merlin-northern, line 39: as long as we are changing this line, perhaps "remort_port"->`remote_port` ?

merlin-northern (approved):
pre-approving. if you are sure about closing Body explicitly in the Dialer, and about the Ticker, we can have it. (the latter should work, the former I have never seem, they always use `_`)

limits/filetransfer/limits.go
lluiscampos: I was unsure about this one too. The lining error is:
```
limits/filetransfer/limits.go:380:3: S1000: should use a simple channel send/receive instead of `select` with a single case (gosimple)
		select {
		^
```

A safer patch that also lints can be:
```
diff --git a/limits/filetransfer/limits.go b/limits/filetransfer/limits.go
index 204ec99..e580f45 100644
--- a/limits/filetransfer/limits.go
+++ b/limits/filetransfer/limits.go
@@ -377,22 +377,21 @@ func updateCounters() {
        tick := time.NewTicker(time.Duration(countersUpdateSleepTimeS) * time.Second)
        defer tick.Stop()
        for counterUpdateRunning {
-               for range tick.C {
-                       txTot := deviceCounters.bytesTransferred
-                       rxTot := deviceCounters.bytesReceived
-                       rx := rxTot - lastRX
-                       tx := txTot - lastTX
-                       lastRX = rxTot
-                       lastTX = txTot
-                       // avg[n+1] = w * Y[n] + (1 - w) avg[n]
-                       avgRX = w1*float64(rx) + _w1*avgRX
-                       avgTX = w1*float64(tx) + _w1*avgTX
-
-                       countersMutex.Lock()
-                       deviceCounters.bytesTransferredAvg1m = avgTX
-                       deviceCounters.bytesReceivedAvg1m = avgRX
-                       countersMutex.Unlock()
-               }
+               <-tick.C
+               txTot := deviceCounters.bytesTransferred
+               rxTot := deviceCounters.bytesReceived
+               rx := rxTot - lastRX
+               tx := txTot - lastTX
+               lastRX = rxTot
+               lastTX = txTot
+               // avg[n+1] = w * Y[n] + (1 - w) avg[n]
+               avgRX = w1*float64(rx) + _w1*avgRX
+               avgTX = w1*float64(tx) + _w1*avgTX
+
+               countersMutex.Lock()
+               deviceCounters.bytesTransferredAvg1m = avgTX
+               deviceCounters.bytesReceivedAvg1m = avgRX
+               countersMutex.Unlock()
        }
 }
 
```

What do you think about this second one? That is basically removing a single case select but blocking the same way it was (which is what I am trying to do with the `for range` too

merlin-northern: makes sense!

session/portforward.go
lluiscampos: Good catch

lluiscampos:
> @mender-test-bot start pipeline shark

Again, "just in cases".

PR #50, opened 2021-12-14
Add couple of info logs to make logging symmetric

lluiscampos:
In both cases will make the whole info level logging more readable.

On start, the last message of regular operations ends with "waiting for
JWT token", so let's inform the user when the token was retrieved and
the connection is up and running.

Similarly, on re-connections, the user is left with "waiting for
reconnect", so inform the user when the re-connection happened.

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -214,6 +214,7 @@ func (d *MenderShellDaemon) messageLoop() (err error) {
 			d.reconnectChan <- e
 			response := <-d.connectionEstChan
 			log.Tracef("messageLoop: got response: %+v", response)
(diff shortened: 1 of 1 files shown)


app/daemon.go
```diff
@@ -453,6 +454,7 @@ func (d *MenderShellDaemon) Run() error {
 			return err
 		}
 		d.authorized = true
+		log.Info("got JWT token, continuing to main loop")
```
merlin-northern, line 457: just for me, could you print the length of the token in the log?

merlin-northern (approved):
only one small request -- feel free to ignore.

merlin-northern:
thank you Sir!

PR #51, opened 2021-12-21
MEN-5290: Several robustness fixes around re-connection

lluiscampos:
* Trigger re-connection when either token or server URL obtained from
  D-Bus signal has changed. Previously, it was enough with having a
  token with "some length" for the code to believe the device was
  authenticated and not trigger a re-connect.
* Abort connecting after 10 retries. Previously it would reconnect
  forever with the same server URL, which will result in a deadlock if
  the url (or the token) changed while re-connecting.

Changelog: Commit


lluiscampos:
@oleorhagen @merlin-northern  There is one TODO we could improve (removing `d.authorized` completely), but I don't have more time before vacation. Unit tests need fixing too.

lluiscampos:
Goes with:
* https://github.com/mendersoftware/meta-mender/pull/1545

lluiscampos:
start pipeline --pr meta-mender/1545 :metal:

lluiscampos:
@mender-test-bot start pipeline --pr meta-mender/1545 :metal:

merlin-northern:
so far I could not find any errors in here. I will give a final review tomorrow.

lluiscampos:
> @mender-test-bot start pipeline --pr meta-mender/1545 metal

:repeat:

lluiscampos:
I force-pushed yesterday night an amend to print the logs of the mock servers, but it didn't help much.

All the logs from the tests are:
```
test_mender-connect.py::TestMenderConnect::test_mender_connect_auth_changes Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Error org.freedesktop.DBus.Error.ServiceUnknown: The name io.mender.AuthenticationManager was not provided by any .service files
FAILED
test_mender-connect.py::TestMenderConnect::test_mender_connect_reconnect Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
Error org.freedesktop.DBus.Error.ServiceUnknown: The name io.mender.AuthenticationManager was not provided by any .service files
FAILED
```

I think that `Pseudo-terminal will not be allocated because stdin is not a terminal` is the right clue to follow. I don't fully understand the implications, but it seems like running the test from GitLab won't accept my strategy of a heredoc with the ssh command and getting the PID. We need to think about a different strategy.

merlin-northern:
> I force-pushed yesterday night an amend to print the logs of the mock servers, but it didn't help much.
> 
> All the logs from the tests are:
> 
> ```
> test_mender-connect.py::TestMenderConnect::test_mender_connect_auth_changes Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Pseudo-terminal will not be allocated because stdin is not a terminal.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Pseudo-terminal will not be allocated because stdin is not a terminal.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Pseudo-terminal will not be allocated because stdin is not a terminal.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Error org.freedesktop.DBus.Error.ServiceUnknown: The name io.mender.AuthenticationManager was not provided by any .service files
> FAILED
> test_mender-connect.py::TestMenderConnect::test_mender_connect_reconnect Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Pseudo-terminal will not be allocated because stdin is not a terminal.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Pseudo-terminal will not be allocated because stdin is not a terminal.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Pseudo-terminal will not be allocated because stdin is not a terminal.
> Warning: Permanently added '[localhost]:8822' (ECDSA) to the list of known hosts.
> Error org.freedesktop.DBus.Error.ServiceUnknown: The name io.mender.AuthenticationManager was not provided by any .service files
> FAILED
> ```
> 
> I think that `Pseudo-terminal will not be allocated because stdin is not a terminal` is the right clue to follow. I don't fully understand the implications, but it seems like running the test from GitLab won't accept my strategy of a heredoc with the ssh command and getting the PID. We need to think about a different strategy.

I have not checked, but if there is ssh in the way, we perhaps could follow the track of `-ttt` option to ssh command and try to do something similar: force to allocate a pseudo tty.

app/daemon.go
```diff
@@ -206,15 +233,17 @@ func (d *MenderShellDaemon) messageLoop() (err error) {
...
-				event: EventReconnectRequest,
+			// nolint:lll
+			// If we used a closed connection means that it has been closed from other goroutine
+			// and a reconnection is ongoing (or done). Just wait for the event and continue
+			// This can happens whe dbusEventLoop detects a change in ServerURL and/or JWT token.
+			// It should be safe to use this string, see:
+			// https://github.com/golang/go/blob/529939072eef730c82333344f321972874758be8/src/net/error_test.go#L502-L507 or about
+			if !strings.Contains(err.Error(), "use of closed network connection") {
+				connectionmanager.Close(ws.ProtoTypeShell)
+				sendConnectReq = true
```
merlin-northern, line 244: I have some doubts here. this was on purpose, although I am not saying it was necessarily right :), do you really want _not_ to close the connection and reconnect on read error?

One possible error scenario returned from Read is: NextReader returning error, gorilla says:
```go
// [...] Errors returned from this method are
// permanent. Once this method returns a non-nil error, all subsequent calls to
// this method return the same error.
```

the other is the ReadAll on the reader. which is the network Read usually. do you think it is safer to close connection and reconnect on any error?

merlin-northern: second comment: 
we set _sendConnectReq = true_ only when there was not a certain error. but we set _waitConnectResp = true_ (lower) always. if waitConnectResp is true the messageLoop waits for message on channel `d.connectionEstChan`, this is only sent on `EventReconnect` in `eventLoop` on `eventChan` which only happens when `needsReconnect && d.authorized` in `dbusEventLoop`, the `d.authorized` is true always if we have a token, and `needsReconnect` is set to true on a message on `reconnectChan` in `dbusEventLoop` or if we detect a change in the token or server url, the message on `reconnectChan` is sent only from messageLoop if `sendConnectReq` is true.

merlin-northern: @oleorhagen while I greatly appreciate the work Lluis did, this code now becomes more and more hard to follow. it was ok at the beginning, but after the recent massive changes on the client side the situation is much different. 

it is really a simple problem:
1. we need one go routine which permanently gets the JWT and server URL over D-Bus. if any of those is empty or changes, it sends reconnect signal.
2. we need to wait for reconnect events, and start reconnecting, each time using `d.jwtToken` and `d.serverURL` (need to be protected by mutex) until we are connected
3. we need to close connection and start reconnecting on errors

what do you think, maybe a small rewrite?

merlin-northern:
@oleorhagen I am unsure how to proceed, see comments above. if I were you, I would rewrite the go routines logic and schema. i.e.: redesign addressing the use case:https://github.com/mendersoftware/mender-connect/pull/51#discussion_r773954831

app/daemon.go
lluiscampos: Be aware that we need to cherry-pick the bugfix to stable branches. If we do a rewrite, consider doing it on a separate commit and cherry-picking only the first one (the one I wrote + potential corrections) to the stable branches.

lluiscampos: @merlin-northern I don't have the mental capacity right now to follow your question. The point is that there are two ways of closing the connection, from `dbusEventLoop` on changes to url or token, or here on remote errors. This `if` is for the former case (we will be getting "closed network connection" error) in which case the connect request has already been sent and we only need to wait for the reconnect.

lluiscampos: > it is really a simple problem:
> 
>     1. we need one go routine which permanently gets the JWT and server URL over D-Bus. if any of those is empty or changes, it sends reconnect signal.
> 
>     2. we need to wait for reconnect events, and start reconnecting, each time using `d.jwtToken` and `d.serverURL` (need to be protected by mutex) until we are connected
> 
>     3. we need to close connection and start reconnecting on errors
> 

The issue is that 1 and 3 will happen at the same time: On changes to the auth parameters, the mender client will close the websocket connection to mender-connect and send a D-Bus signal with the new parameters. That is exactly why we have the catch for "closed network connection" error, as this indicates that the D-Bus loop has gotten the event before the message loop.

app/daemon.go
```diff
@@ -277,7 +277,7 @@ func (d *MenderShellDaemon) gotAuthToken(p []dbus.SignalParams, needsReconnect b
 			shellsCount, sessionsCount, err := session.MenderSessionTerminateAll()
 			if err == nil {
 				log.Infof("dbusEventLoop terminated %d sessions, %d shells",
-					shellsCount, sessionsCount)
+					sessionsCount, shellsCount)
```
oleorhagen, line 280: Unrelated to this commit, no?

config/paths.go
```diff
@@ -39,7 +39,7 @@ var (
 	DefaultDebug = false
 	DefaultTrace = false
 
-	MaxReconnectAttempts             = uint(0) // 0 means to reconnect forever
+	MaxReconnectAttempts             = uint(10)
```
oleorhagen, line 42: Question: 
Why this change?

connectionmanager/connectionmanager.go
```diff
@@ -132,7 +132,13 @@ func connect(
 		mutex:      &sync.Mutex{},
 	}
 
-	log.Tracef("connectionmanager connect returning")
+	// TODO: to be tested, did this really fix the deadlock?
+	select {
+	case <-cancelReconnectChan[proto]:
+		log.Trace("connectionmanager drained cancelReconnectChan")
```
oleorhagen, line 138: Question:
It's not necessarily drained though, just one element taken out?

connectionmanager/connectionmanager.go
```diff
@@ -211,9 +217,11 @@ func setReconnecting(proto ws.ProtoType, v bool) bool {
 	return v
 }
 
-func CancelReconnection(proto ws.ProtoType) {
+func CancelReconnection(proto ws.ProtoType) bool {
 	maxWaitSeconds := 8
-	cancelReconnectChan[proto] <- true
+	go func() {
+		cancelReconnectChan[proto] <- true
```
oleorhagen, line 223: Suggestion:
What about just giving the channel a buffer size?

oleorhagen (approved):
I'm not even gonna pretend that I got under the surface on this one, there is a lot of moving pieces here.

I think the approach is sensible though, although the code is hard to decipher. Only superficial comments added.

Maybe you could actually give me a guide through the code f2f tomorrow?

app/daemon.go
lluiscampos: Correct. I will split it out.

config/paths.go
lluiscampos: From the commit message:
```
Abort connecting after 10 retries. Previously it would reconnect
forever with the same server URL, which will result in a deadlock if
the url (or the token) changed while re-connecting.
```

connectionmanager/connectionmanager.go
lluiscampos: This I am not very sure about. Only one element is taken out but is only one place where elements are put in, so it is "drained" because is unbuffered.

I think that having moved the sender (line 222) to a goroutine is enough and we can remove this part here. If it exists with a pending element in the channel here it will only be a problem next time the reconnection is attempted to be cancelled, as it will timeout there.

But it should be harmful anyway, so I decided to leave with trace logging.

connectionmanager/connectionmanager.go
lluiscampos: How will a buffered channel prevent the deadlock? Won't it be the same problem? The base problem is that during the reconnection there is a for loop (previously forever, now up to 10 attempts) where this channel might be read but the loop can also return on errors or on success without reading the channel (hence this "draining" I also wrote there, line 138).

config/paths.go
oleorhagen: Shame on me :see_no_evil:

lluiscampos:
> > @mender-test-bot start pipeline --pr meta-mender/1545 metal
> 
> repeat

:dancing_men: :dancing_women:

lluiscampos:
> > > @mender-test-bot start pipeline --pr meta-mender/1545

ping

lluiscampos:
> > > > @mender-test-bot start pipeline --pr meta-mender/1545

One more time :balloon:

lluiscampos:
> > > > > @mender-test-bot start pipeline --pr meta-mender/1545

:roll_eyes:

lluiscampos:
cherry-pick to:
* 3.1.x
* 2.6.x

PR #68, opened 2022-05-31
chore: added missing dependabot config

mzedel:
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - commit-message:
+      prefix: "chore"
+    directory: /
+    package-ecosystem: gomod
+    schedule:
+      interval: monthly


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - commit-message:
+      prefix: "chore"
+    directory: /
+    package-ecosystem: gomod
+    schedule:
+      interval: monthly
```
merlin-northern, line 8: is there a reason for this being monthly, not weekly? (mender-client has weekly) (just trying to be picky) otherwise looks good, approving not to block.

merlin-northern (approved):

.github/dependabot.yml
mzedel: yes, we decided to move to monthly updates to reduce the frequency, since we already have so many open PRs to get to with some packages just releasing very frequently - I'll align the other repos eventually...

mzedel: also... funny to see the unittest coverage increase...

merlin-northern: thank you sorry for the noise.

merlin-northern: > also... funny to see the unittest coverage increase...

well, I am very sorry, but you will just have to add some tests. `time.Sleep(1.5*time.Month)` and see if dependabot shows something :>

PR #80, opened 2022-10-18
go fmt: required formatting.

merlin-northern:
chore: go fmt: required formatting.
Ticket: None

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


connection/connection.go
```diff
@@ -42,7 +42,7 @@ type Connection struct {
 	done chan bool
 }
 
-//Websocket connection routine. setup the ping-pong and connection settings
+// Websocket connection routine. setup the ping-pong and connection settings
```
merlin-northern, line 45: these we have to have, right?

kacf: Yes, those are fine.

kacf (approved):

merlin-northern:
formatting, not running the pipeline.

PR #89, opened 2023-01-09
Fix limits test user name mixup

andhe:
currentUser is supposed to have the current uid's username, not the descriptive field of the user (aka gecos).
Atleast that's what it gets compared against eventually.

The following two tests are affected and could sometimes give false-positives (since there's an explicit check if the user name is the empty string and then return true, which is true in some build scenarios when gecos is empty) but could also fail when gecos contained the actual users full name.

task-0:     --- FAIL: TestPermit_UploadFile/file_exists_allowed_to_overwrite_owner_match (0.00s)

task-0:     --- FAIL: TestPermit_DownloadFile/file_owner_match (0.00s)

### DIFF
--- a/limits/filetransfer/limits_test.go
+++ b/limits/filetransfer/limits_test.go
@@ -388,7 +388,7 @@ func TestPermit_DownloadFile(t *testing.T) {
 	if u == nil {
 		t.Fatal("cant get current user")
 	}
-	currentUser := u.Name
+	currentUser := u.Username
 	currentGroup, _ := user.LookupGroupId(u.Gid)
 	if currentGroup == nil {
 		t.Fatal("cant get current group")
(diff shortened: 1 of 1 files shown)


lluiscampos (changes requested):
Good catch! Thank you @andhe 

The fix looks good :+1:  We need to handle the same two formalities than in https://github.com/mendersoftware/mender-connect/pull/90#pullrequestreview-1261265076 : prefix `test: ` in the commit title and update copyright here in the source code file.

Let me know if you want me to amend this for you!

andhe:
Hi @lluiscampos 

Thanks for looking at this. Please make any changes as you see fit to take it to the next step.

lluiscampos:
Merged in #91.

Thank you @andhe!

PR #90, opened 2023-01-10
Fix racing connection tests

andhe:
The listening tcp port needs to be up before we connect or the client will end up with "connection refused".

When listen is done in a (go) thread, there's no guarantee the thread even gets to execute before the test moves on to do the connect.

Insert a sleep after thread creation in tests that has been seen to be flaky (on debian build infrastructure), hopefully long enough that the thread doing the listening gets a chance to run first.

The TestPortForwardHandlerSuccessfulConnection was reproduced locally (reliably) and the change confirmed to fix the issue there. The TestReconnect is an approximation that it's the same/similar issue and hopefully the sleep will fix it.

Relevant build logs from debian infrastructure showing the problems: https://buildd.debian.org/status/fetch.php?pkg=mender-connect&arch=ppc64el&ver=2.1.0%2Bds1-3&stamp=1673302707&raw=0 https://buildd.debian.org/status/fetch.php?pkg=mender-connect&arch=armel&ver=2.1.0%2Bds1-3&stamp=1673345986&raw=0 https://buildd.debian.org/status/fetch.php?pkg=mender-connect&arch=armhf&ver=2.1.0%2Bds1-3&stamp=1673322854&raw=0

Signed-off-by: Andreas Henriksson <andreas@fatal.se>

### DIFF
--- a/connectionmanager/connectionmanager_test.go
+++ b/connectionmanager/connectionmanager_test.go
@@ -123,6 +123,8 @@ func TestReconnect(t *testing.T) {
 		_ = srv.Shutdown(context.Background())
 	}()
 
+	time.Sleep(2000 * time.Millisecond)
+
 	_ = Close(ws.ProtoTypeShell)
 
 	ctx := context.Background()
--- a/session/portforward_test.go
+++ b/session/portforward_test.go
@@ -218,6 +218,8 @@ func TestPortForwardHandlerSuccessfulConnection(t *testing.T) {
 		}
 	}(tcpPort)
 
+	time.Sleep(2000 * time.Millisecond)
+
 	// c1: new
 	protocol := wspf.PortForwardProtocol(wspf.PortForwardProtocolTCP)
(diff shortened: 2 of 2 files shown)


lluiscampos (changes requested):
Hi @andhe and thank you for following up on this.

The change itself looks good, we have not seen these errors on our CI but there it no inconvenience in making the tests Debian CI friendly :+1: 

Two formalities that we need to fix before merge:
* The commit needs to adhere to our conventional commits convention (doc [here](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#structuring-your-commits)). For this case, it is enough with just prefixing `test: ` on your commit title.
* Update the Copyright year to `2023` for the two source files that were modified in this PR: `connectionmanager_test.go` and `portforward_test.go`.
```

Thank you!

lluiscampos:
Merged in #92.  Thanks @andhe :+1:

PR #100, opened 2023-05-11
2.0.x unit tests fixes (cherry picks)

merlin-northern:


merlin-northern:
cherry-picks to tests only, not running the full pipeline.

PR #101, opened 2023-06-22
ci: Update unit tests job to the latest template

lluiscampos:
So that it produces a junit report that can be parsed by CI.


mzedel (approved):
🙏 thank you for the alignment work! even helped to improve the coverage 🚀

PR #111, opened 2023-11-01
chore: Change the systemd service file to the new mender-authd service

oleorhagen:
I thought about versioning this for a little, and I am not really sure.

I decided on no version bump for this, as it is an external requirement to the software. Tell me if I am in the wrong here

### DIFF
--- a/support/mender-connect.service
+++ b/support/mender-connect.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=Mender Connect service
 Wants=network-online.target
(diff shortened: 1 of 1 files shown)


lluiscampos:
As discussed f2f, this needs to be backwards compatible. The fix probably needs to live in `meta-mender`

oleorhagen:
Since we _have_ to be backwards compatible :( here, this will not work.

oleorhagen:
lazy 0 - world 1

PR #112, opened 2023-11-02
MEN-6827: Support new CLI interface of the mender client

lluiscampos:
Changelog: In Mender client v4.0, the `mender` binary has been split into `mender-update` and `mender-auth`. Modify `mender-connect` to call the right binary to trigger `check-update` and `send-inventory` Ticket: MEN-6671

### DIFF
--- a/session/menderclient.go
+++ b/session/menderclient.go
@@ -15,6 +15,7 @@
 package session
 
 import (
+	"os"
 	"os/exec"
 
 	"github.com/mendersoftware/go-lib-micro/ws"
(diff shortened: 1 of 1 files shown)


lluiscampos:
Please review also the commit classification,

In my opinion it is a fix from the `mender-connect` point of view, which does not honor a minor release but that needs to be cherry-picked to all supported versions of `mender-connect` (so that they will work with latest client).

kacf (approved):
Agree with the classification.

lluiscampos:
* 2.1.x
* 2.0.x

session/menderclient.go
```diff
@@ -35,6 +36,18 @@ func MenderClient() Constructor {
 	return func() SessionHandler { return f }
 }
 
+func getMenderClientBinaryPath() (string, error) {
+	candidates := []string{"/usr/bin/mender-update", "/usr/bin/mender"}
```
oleorhagen, line 40: Q: Is it not better to check if they are in the `PATH` ? :thinking:

lluiscampos: Maybe you are right. Something like [this](https://github.com/mendersoftware/mender-artifact/blob/master/utils/binpath.go#L29)?

Is it okey if I push this a bit down the road? I make note as a low priority.

oleorhagen: Absolutely - it is not a regression in any way :)

oleorhagen: > Maybe you are right. Something like [this](https://github.com/mendersoftware/mender-artifact/blob/master/utils/binpath.go#L29)?

Yes :)

PR #118, opened 2023-11-28
feat: added back-off in connectionmanager

danielskinstad:
Exponentially backs off until max (5 minutes) is reached. It will then keep retrying every 5 minutes.

Changelog: Title

Ticket: MEN-5782

### DIFF
--- a/connectionmanager/connectionmanager.go
+++ b/connectionmanager/connectionmanager.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2023 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -59,72 +59,58 @@ type ProtocolHandler struct {
 
 var handlersByTypeMutex = &sync.Mutex{}
 var handlersByType = map[ws.ProtoType]*ProtocolHandler{}
-var reconnectIntervalSeconds = 5
 var DefaultPingWait = time.Minute
 
 func GetWriteTimeout() time.Duration {
 	return writeWait
 }
 
-func SetReconnectIntervalSeconds(i int) {
-	reconnectIntervalSeconds = i
+var a *expBackoff = &expBackoff{
+	maxInterval:  60 * time.Minute,
+	attempts:     0,
+	exceededMax:  false,
+	maxBackoff:   120 * time.Minute,
+	smallestUnit: time.Minute,
 }
 
 func connect(
 	proto ws.ProtoType,
 	serverUrl, connectUrl, token string,
-	retries uint,
 	ctx context.Context,
 ) error {
 	parsedUrl, err := url.Parse(serverUrl)
 	if err != nil {
 		return err
 	}
-
 	scheme := getWebSocketScheme(parsedUrl.Scheme)
 	u := url.URL{Scheme: scheme, Host: parsedUrl.Host, Path: connectUrl}
 
 	cancelReconnectChan[proto] = make(chan bool)
 	var c *connection.Connection
-	var i uint = 0
 	defer func() {
 		setReconnecting(proto, false)
 	}()
+
 	setReconnecting(proto, true)
-	for IsReconnecting(proto) {
-		i++
-		c, err = connection.NewConnection(u, token, writeWait, maxMessageSize, DefaultPingWait)
-		if err != nil || c == nil {
-			if retries == 0 || i < retries {
-				if err == nil {
-					err = errors.New(
-						"unknown error: connection was nil but no error provided by" +
-							" connection.NewConnection",
-					)
-				}
-				log.Errorf("connection manager failed to connect to %s%s: %s; "+
-					"reconnecting in %ds (try %d/%d); len(token)=%d", serverUrl, connectUrl,
-					err.Error(), reconnectIntervalSeconds, i, retries, len(token))
-				select {
-				case cancelFlag := <-cancelReconnectChan[proto]:
-					log.Tracef("connectionmanager connect got cancelFlag=%+v", cancelFlag)
-					if cancelFlag {
-						return nil
-					}
-				case <-ctx.Done():
-					return nil
-				case <-time.After(time.Second * time.Duration(reconnectIntervalSeconds)):
-					break
-				}
-				continue
-			} else if i >= retries {
-				return ErrConnectionRetriesExhausted
-			}
-			return err
-		} else {
-			break
-		}
+	if err := a.WaitForBackoff(ctx, proto); err != nil {
+		log.Errorf("error in WaitForBackoff: %s", err)
+		return err
 	}
+	c, err = connection.NewConnection(u, token, writeWait, maxMessageSize, DefaultPingWait)
+	if err != nil || c == nil {
+		if err == nil {
+			err = errors.New(
+				"unknown error: connection was nil but no error provided by" +
+					" connection.NewConnection",
+			)
+		}
+		log.Errorf("connection manager failed to connect to %s%s: %s; "+
+			"len(token)=%d", serverUrl, connectUrl,
+			err.Error(), len(token))
+		return err
+	}
+
+	a.resetBackoff()
 
 	handlersByType[proto] = &ProtocolHandler{
 		proto:      proto,
@@ -145,7 +131,6 @@ func connect(
 func Connect(
 	proto ws.ProtoType,
 	serverUrl, connectUrl, token string,
-	retries uint,
 	ctx context.Context,
 ) error {
 	handlersByTypeMutex.Lock()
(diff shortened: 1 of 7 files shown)


alfrunes (changes requested):
Great work getting your hands dirty with some Go programming, starting with `mender-connect` is pretty brave :smile_cat:
First off, when it comes to the exponential back-off, I believe there should only be exponential back-off when trying to establish the websocket connection as this is the only path that directly interacts with the server. Your implementation is adding delay one step further out - interacting with the `mender` client on the device (using the dbus api). So I would first move the logic to the `connectionmanager` package.
Second, when it comes to the algorithm itself, it seems a bit arbitrary. Imagine millions of devices connected to a server and then a transient error causes all devices to loose connection. Now, all clients will in unison try to reconnect to the server which in turn significantly increase the burden for the server to recover. So I suggest we add a little bit of randomness delay on top of the exponential backoff, as well as increase the maximum interval. Ideally I would prefer if there instead were a cutoff interval where the algorithm changes to increase the interval linearly. For example, do exponential random backoff up to 1 hour. Then, for each attempt, increase the interval by 1-2 minutes (randomly chosen).

connectionmanager/connectionmanager.go
```diff
@@ -59,72 +59,67 @@ type ProtocolHandler struct {
...
-					}
-				case <-ctx.Done():
-					return nil
-				case <-time.After(time.Second * time.Duration(reconnectIntervalSeconds)):
-					break
-				}
-				continue
-			} else if i >= retries {
-				return ErrConnectionRetriesExhausted
+	c, err = connection.NewConnection(u, token, writeWait, maxMessageSize, DefaultPingWait)
```
alfrunes, line 95: The back-off should protect `connection.NewConnection`. For example:
```go
	if err := a.Backoff(ctx); err != nil {
		return err
	}
	c, err = connection.NewConnection(u, token, writeWait, maxMessageSize, DefaultPingWait)
```
Where `a.WaitForBackoff` would look something like:
```go
func (a *expBackoff) Backoff(ctx context.Context, proto ws.ProtoType) error {
	a.attempts++
	if a.attempts <= 1 {
		return nil
	}
	intrval := a.GetExponentialBackoffTime()
	select {
	case <-time.After(a.GetExponentialBackoffTime()):
	case <-cancelReconnectChan[proto]:
		return context.Canceled
	case <-ctx.Done():
		return ctx.Err()
	}
	return nil
}
```

connectionmanager/connectionmanager.go
```diff
@@ -59,72 +59,67 @@ type ProtocolHandler struct {
...
+		case cancelFlag := <-cancelReconnectChan[proto]:
+			log.Tracef("connectionmanager connect got cancelFlag=%+v", cancelFlag)
+			if cancelFlag {
+				return nil
 			}
+		case <-ctx.Done():
+			return nil
+		// Waits for the interval calculated by GetExponentialBackoffTime
+		case <-time.After(intvl):
 			return err
```
alfrunes, line 119: Waiting to return the already available error is a waste of time. The waiting should happen before the call to establish the connection.

connectionmanager/exponentialbackoff.go
```diff
@@ -0,0 +1,60 @@
...
+	// Smallest backoff sleep time
+	smallestUnit time.Duration
+}
+
+func (a *expBackoff) GetExponentialBackoffTime() time.Duration {
+	var nextInterval time.Duration
+	const perIntervalAttempts = 3
+
+	interval := 1 * a.smallestUnit
+	nextInterval = interval << (a.attempts / perIntervalAttempts)
```
alfrunes, line 38: The `perIntervalAttempts` seems a bit arbitrary. I know this comes from the mender client, but I do not understand the purpose of such a mechanism.

alfrunes (changes requested):
The back-off needs to happen before the call to `connection.NewConnection`.
I know we also talked about the removal of retries, but now I'm not sure. I don't know the intricacies about the dbus API, I'd like to get some input from @merlin-northern or @lluiscampos about this.

merlin-northern:
> The back-off needs to happen before the call to `connection.NewConnection`. I know we also talked about the removal of retries, but now I'm not sure. I don't know the intricacies about the dbus API, I'd like to get some input from @merlin-northern or @lluiscampos about this.

sorry I am a bit late for the party. @alfrunes what shall I do? do you want me to give a full review? the dbus API: without reading the current code, and as far as I remember there is a separete go routine which monitors for the dbus events. in general what you have eto take into account with the back off is the fac that you can loose authorization at any given moment. let me know how I can be os assistance.

connectionmanager/exponentialbackoff.go
```diff
@@ -0,0 +1,60 @@
...
+	"math/rand"
+	"time"
+)
+
+type expBackoff struct {
+	attempts    int
+	exceededMax bool
+	// The maximum interval before it starts increasing linearly
+	maxInterval time.Duration
+	// Retry time does not increaseb beyond maxBackoff
```
lluiscampos, line 27:
```suggestion
	// Retry time does not increase beyond maxBackoff
```

lluiscampos (changes requested):
> The back-off needs to happen before the call to `connection.NewConnection`. 

I agree

> I know we also talked about the removal of retries, but now I'm not sure. I don't know the intricacies about the dbus API, I'd like to get some input from @merlin-northern or @lluiscampos about this.

As far as I can see, this should be fine. From the `dbusEventLoop` the ongoing (re)connection is cancelled* when the server url or the token change, so it will exit back to the main context in `eventLoop` and then start the new connection attempt.

(*) It is not very intuitive, but in `processJwtTokenStateChange` we call `connectionmanager.Close` which in turn will send the cancel through `cancelReconnectChan[proto]`.

Now, one could argue that this was the case before and that `MaxReconnectAttempts` did not have any reason to exist. So I might be missing something :eyes: Am I?

If you @alfrunes nor @merlin-northern aren't confident with my theory, then I suggest that we re-introduce `MaxReconnectAttempts` and modify the backoff algorithm of this PR to stop after such many attempts, this way it will still behave as it used to wrt eventually going back to `eventLoop` and starting all over (re)connection.

alfrunes:
> If you @alfrunes nor @merlin-northern aren't confident with my theory, then I suggest that we re-introduce `MaxReconnectAttempts` and modify the backoff algorithm of this PR to stop after such many attempts, this way it will still behave as it used to wrt eventually going back to `eventLoop` and starting all over (re)connection.

To be honest, I'm not very familiar with the connection management and dbus stuff, but I gave it a shot creating the simplest wrapper around the websocket interface that I could come up with: https://github.com/mendersoftware/mender-connect/pull/133

connectionmanager/connectionmanager.go
```diff
@@ -59,72 +59,67 @@ type ProtocolHandler struct {
...
-	reconnectIntervalSeconds = i
+var a *expBackoff = &expBackoff{
+	maxInterval:  60 * time.Minute,
+	attempts:     0,
+	exceededMax:  false,
+	maxBackoff:   120 * time.Minute,
+	smallestUnit: time.Minute,
 }
 
 func connect(
```
merlin-northern, line 76: @alfrunes  @danielskinstad @lluiscampos I leav the details of the review for you. what I verified is: this (the `connect` function) is the correct place to introduce the backoff -- the reconnect mechanism via events is independent (mostly) from the dbus events.

connectionmanager/exponentialbackoff.go
```diff
@@ -0,0 +1,96 @@
...
+	case <-cancelReconnectChan[proto]:
+		return context.Canceled
+	case <-ctx.Done():
+		return ctx.Err()
+	case cancelFlag := <-cancelReconnectChan[proto]:
+		log.Tracef("connectionmanager connect got cancelFlag=%+v", cancelFlag)
+		a.cancelFlag = cancelFlag
+		if cancelFlag {
+			return nil
+		}
```
lluiscampos, line 88: This is not necessary as it will `return nil` anyway on the next statement

connectionmanager/exponentialbackoff.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (a *expBackoff) Backoff(ctx context.Context, proto ws.ProtoType) error {
+	a.attempts++
+	a.interval = a.GetExponentialBackoffTime()
+
+	if a.attempts <= 1 {
+		return nil
+	}
+
+	log.Errorf("connectionmanager backoff: retrying in %s", a.interval)
```
lluiscampos, line 75: This is not an error, I suggest using an `Info` log entry here.

lluiscampos (approved):
I think this is coming together pretty nicely now! I commented on few things, but overall looks good so I am pre-approving it.

With regards to your Git history,
* The two commits in this PR should really be one. We are not very consistent in the team, but the principle should be that each commit is "buildable" - meaning if you change function you should fix the tests in the same commit.
* The title for your commits are very long. Please have a look at [this source](https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message) (and, specifically, the vim configuration :wink: ). This is one of your first PRs and I know you have improved the messaging since, but having vim highlighting the syntax will help you write better commit messages when amending too :smiley:

danielskinstad:

danielskinstad:
@mender-test-bot start pipeline --pr meta-mender/2080

alfrunes (commented):
*Review dismissed

alfrunes (approved):
Unblocking my blocking review.

connectionmanager/connectionmanager.go
merlin-northern: I have total and uttermost trust in my team mates. proceed.

PR #119, opened 2023-11-28
MEN-6858: Update sytemd service for the new client service

lluiscampos:
Changelog: `mender-connect` systemd service can use either the old
`mender-client` service or the new `mender-authd` service, which is the
name of the authentication daemon of new Mender client. They are both
kept as `Wants` and not `Require` so that `mender-connect` can run with
either.

### DIFF
--- a/support/mender-connect.service
+++ b/support/mender-connect.service
@@ -1,8 +1,7 @@
 [Unit]
 Description=Mender Connect service
-Wants=network-online.target
-After=systemd-resolved.service network-online.target mender-client.service
-Requires=mender-client.service
+Wants=network-online.target mender-client.service mender-authd.service
+After=systemd-resolved.service network-online.target mender-client.service mender-authd.service
 
 [Service]
 Type=idle


support/mender-connect.service
```diff
@@ -2,7 +2,7 @@
 Description=Mender Connect service
 Wants=network-online.target
 After=systemd-resolved.service network-online.target mender-client.service
-Requires=mender-client.service
+Requires=mender-authd.service
```
kacf, line 5: I think we should re-think this approach. It should not be mandatory to upgrade to mender 4.0.0 to keep using mender-connect. But I don't think it's possible to achieve that with `Requires`, because there is no OR.

So my proposal is: Switch to `Wants=`, which is weaker, but still strong enough for our purposes, I believe. Then you can list both services, and it won't fail to start even though one will always be missing. If both are missing, it will still start, and won't work, but I don't think this is significantly worse than not starting.

support/mender-connect.service
```diff
@@ -2,7 +2,7 @@
 Description=Mender Connect service
 Wants=network-online.target
 After=systemd-resolved.service network-online.target mender-client.service
```
kacf, line 4: Remember to list mender-authd here as well.

kacf (changes requested):

support/mender-connect.service
lluiscampos: Thank you.

I actually considered this for a long while, but eventually I thought that it should have good reasons to be already in `Requires` (for example to be stopped if `mender-client` is stopped).

But once you bring it up, and in light of being defensive from `mender-connect` point of view, it makes sense to switch to `Wants`.

Note also that I have now marked the commit as `fix` and not `feat`, so that we can (and will) cherry-pick to stable branches.

kacf (approved):

lluiscampos:
@mender-test-bot cherry-pick to 2.1.x, 2.0.x

lluiscampos:
* 2.1.x
* 2.0.x

PR #128, opened 2024-01-18
chore: removed the text "mender-connect version" from --version outpu…

danielskinstad:
…t to be consistent

Changelog: --version no longer outputs the text "mender-connect version".

Ticket: MEN-6965

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -70,6 +70,13 @@ func SetupCLI(args []string) error {
 		},
 	}
 
+	cli.VersionPrinter = func(c *cli.Context) {
+		err := config.ShowVersionCLI(c)
+		if err != nil {
+			return
+		}
+	}
+
 	return app.Run(args)
 }
 


lluiscampos (changes requested):
The code looks good.

However you are doing two things in this commit: to remove the `version` command and to modify the output for the `--version` flag. This could have very well been two separate commits.

For the first part, we are technically speaking removing a command (the `version` command) and it is a breaking change, which will require a major version bump.

Either:
* Drop the removal of `version` command.
* Separate the commits and mark the removal of `version` command as a [breaking change](https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md#notes).
    * @oleorhagen I think it is an overkill, but I'd rather be technically correct wrt semantic version that run into trouble in the future because someone out there did use this command... What do you think?

oleorhagen:
> The code looks good.
> 
> However you are doing two things in this commit: to remove the `version` command and to modify the output for the `--version` flag. This could have very well been two separate commits.
> 
> For the first part, we are technically speaking removing a command (the `version` command) and it is a breaking change, which will require a major version bump.
> 
> Either:
> 
>     * Drop the removal of `version` command.
> 
>     * Separate the commits and mark the removal of `version` command as a [breaking change](https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md#notes).
>       
>       * @oleorhagen I think it is an overkill, but I'd rather be technically correct wrt semantic version that run into trouble in the future because someone out there did use this command... What do you think?

Yes, I agree. I mean, this could've even broken some of our stuff

oleorhagen:
So, yeah. I mean. I vote for keeping the command. That is option 1.

lluiscampos (approved):

danielskinstad:

lluiscampos:
The failing test is a known and unrelated issue. Merging

PR #130, opened 2024-01-26
MEN-6888: Stop shell after health check fails

lluiscampos:
`mender-connect` has two levels of ping/pong, the standard websockets one and a custom `shell` protocol one. On the later, when the peer does not respond to the ping request the shell session, the session is marked as expired but there was no default mechanism to stop the shell and get ready for new shells.

The symptom of this bug is that after temporally lose of connectivity, `mender-connect` will still refuse to open new shells after the connectivity is restored.

It was possible to configure `mender-connect` to stop shells from expired sessions. See changelog entry below.

Ticket: MEN-6888
Changelog: Stop shell after the ping/pong health check fails, so that the `mender-connect` can recover after disruption of connectivity. For already released versions of `mender-connect` the workaround is to force the stop of shells for expired sessions adding to the configuration options the fields:
```
  "Sessions": {
    "StopExpired": true,
    "ExpireAfterIdle": 600
  }
```
Setting a limit for expiration high enough (10 minutes, above) so that it doesn't interfere with regular operations.
See https://docs.mender.io/add-ons/mender-connect#remote-terminal-configuration

### DIFF
--- a/session/shell.go
+++ b/session/shell.go
@@ -212,6 +212,18 @@ func MenderShellSessionsGetByUserId(userId string) []*MenderShellSession {
 	}
 }
 
+func MenderShellStopById(sessionId string) error {
+	s := MenderShellSessionGetById(sessionId)
+	if s.shell == nil {
+		return ErrSessionNotFound
+	}
+	e := s.StopShell()
+	if e != nil && procps.ProcessExists(s.shellPid) {
+		return e
+	}
+	return MenderShellDeleteById(sessionId)
+}
+
 func MenderShellStopByUserId(userId string) (count uint, err error) {
 	a := sessionsByUserIdMap[userId]
 	log.Debugf("stopping all shells of user %s.", userId)
@@ -396,6 +408,12 @@ func (s *MenderShellSession) healthcheck() {
 			if s.healthcheckTimeout.Before(time.Now()) {
 				log.Errorf("session %s, health check failed, connection with the client lost", s.id)
 				s.expiresAt = time.Now()
+				err := MenderShellStopById(s.id)
+				if err != nil {
+					log.Errorf("session %s, error stopping shell %s", s.id, err.Error())
+				} else {
+					log.Infof("session %s successfully stopped", s.id)
+				}
 				return
 			}
 		case <-time.After(time.Until(nextHealthcheckPing)):


lluiscampos:
@mender-test-bot start pipeline --pr integration/2558

lluiscampos:
It is tested at:
* https://github.com/mendersoftware/integration/pull/2558

In the acceptance criteria of the task we require to have a *unit* test for this, but I didn't find a simple way to test it :see_no_evil: @alfrunes @merlin-northern maybe you guys can guide me? Or do you think it is enough with the integration test?

oleorhagen (approved):
Nice!

What happens on the other end (the GUI) when this happens? Maybe we should do a sanity check there as well?

oleorhagen:
> > What happens on the other end (the GUI) when this happens? Maybe we should do a sanity check there as well?
> 
> @oleorhagen As far as I can see, the shell ping/pong is always initiated on the device side and responded by the GUI - in other words: the device can only send shell ping messages and can only receive shell pong messages.
> 
> I am not able to read the GUI code, but I experimented by making the device side ignore the pong from GUI and close the connection. In this case the GUI still thinks that the session is active, but when sending any message on the console you get an error. Closing and opening the console re-establishes the connection (a new session, really).

Alright, sweet. 

Thanks for looking into it!

merlin-northern (commented):
isnt it enough to just switch the configuration of expired session to on by default?

merlin-northern:
> It is tested at:
> 
> * [MEN-6888: Re-enable (and modify) `test_in_poor_network_environment` integration#2558](https://github.com/mendersoftware/integration/pull/2558)
> 
> In the acceptance criteria of the task we require to have a _unit_ test for this, but I didn't find a simple way to test it 🙈 @alfrunes @merlin-northern maybe you guys can guide me? Or do you think it is enough with the integration test?

I would say it is ok if we have integration test for it. if we hadnt I would probably vote for a unit test (our units in here do lots more than just unit testing).

lluiscampos:
> isnt it enough to just switch the configuration of expired session to on by default?

It would need to be switched on *and* to have a default value for expiration.

I considered this. But that is a behavioral change rather than a bug fix (it will actually expire idle sessions after the default time) and I think it will be unexpected for users that might be running long sessions.

lluiscampos:
* 2.2.x
* 2.1.x

lluiscampos:
* 2.2.x
* 2.1.x

PR #133, opened 2024-01-30
feat: Add exponential random backoff to websocket interface

alfrunes:
The application applies exponential backoff for every attempt to connect to the websocket API until a successful connection is made.

The back-off algorithm uses exponential factor 2 and jitter of 1 second, starting from an initial back-off of 1 second.

Changelog: Commit
Ticket: MEN-5782

### DIFF
--- a/connectionmanager/connectionmanager.go
+++ b/connectionmanager/connectionmanager.go
@@ -93,7 +93,8 @@ func connect(
 	setReconnecting(proto, true)
 	for IsReconnecting(proto) {
 		i++
-		c, err = connection.NewConnection(u, token, writeWait, maxMessageSize, DefaultPingWait)
+		c, err = connection.NewConnection(ctx, u, token,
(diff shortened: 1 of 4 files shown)


connectionmanager/connectionmanager.go
```diff
@@ -93,7 +93,8 @@ func connect(
 	setReconnecting(proto, true)
 	for IsReconnecting(proto) {
 		i++
-		c, err = connection.NewConnection(u, token, writeWait, maxMessageSize, DefaultPingWait)
+		c, err = connection.NewConnection(ctx, u, token,
+			writeWait, maxMessageSize, DefaultPingWait)
```
merlin-northern, line 97: I verified that this is the correct place to introduce the back-off. I also verified that we need it in this (very elegant) form or the other PR (also not bad).

alfrunes: Oops... I forgot to push my latest amendments to this PR.

PR #134, opened 2024-02-29
fix: Fix to make mender-connect request a new JWT token on its own, so its not dependent on mender-update to do so

SebOpsahl:
Changelog: Fix to make `mender-connect` request a new JWT token on its own, so its not dependent on `mender-update` to do so. Previously, `mender-connect` relied the Mender client to request new JWT tokens. Since the split of mender into `mender-update` and `mender-auth`, it is theoretically possible to run mender-connect without mender-update, in which case it wouldn't request JWT tokens without this fix.

Ticket: MEN-6877

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -319,21 +319,28 @@ func (d *MenderShellDaemon) dbusEventLoop(client mender.AuthClient) {
 
 		p, err := client.WaitForJwtTokenStateChange()
 		log.Tracef("dbusEventLoop: WaitForJwtTokenStateChange %v, err %v", p, err)
+
 		if len(p) > 1 &&
 			p[0].ParamType == dbus.GDBusTypeString &&
 			p[1].ParamType == dbus.GDBusTypeString {
+
 			token := p[0].ParamData.(string)
 			serverURL := p[1].ParamData.(string)
+
 			d.processJwtTokenStateChange(token, serverURL)
 			if len(token) > 0 && len(serverURL) > 0 {
 				log.Tracef("dbusEventLoop: got a token len=%d, ServerURL=%s", len(token), serverURL)
 				if token != d.serverJwt || serverURL != d.serverUrl {
-					log.Debugf(
-						"dbusEventLoop: new token or ServerURL, reconnecting. len=%d, ServerURL=%s",
-						len(token),
-						serverURL,
-					)
-					needsReconnect = true
+					e := MenderShellDaemonEvent{
+						event: EventReconnect,
+						data:  token,
+						id:    "(dbusEventLoop)",
+					}
+					d.serverJwt = token
+					d.serverUrl = serverURL
+					d.postEvent(e)
+					log.Tracef("(dbusEventLoop) posting Event: %s", e.event)
+					needsReconnect = false
 
 					// If the server (Mender client proxy) closed the connection, it is likely that
 					// both messageLoop is asking for a reconnection and we got JwtTokenStateChange
@@ -341,28 +348,18 @@ func (d *MenderShellDaemon) dbusEventLoop(client mender.AuthClient) {
 					if d.needsReconnect() {
 						log.Debug("dbusEventLoop: drained reconnect req channel")
 					}
-
 				}
-				// TODO: moving these assignments one scope up would make d.authorized redundant...
-				d.serverJwt = token
-				d.serverUrl = serverURL
 			}
 		}
-		if needsReconnect && d.authorized {
-			jwtToken, serverURL, _ := client.GetJWTToken()
-			e := MenderShellDaemonEvent{
-				event: EventReconnect,
-				data:  jwtToken,
-				id:    "(dbusEventLoop)",
+
+		if needsReconnect {
+			log.Infof("Within needsReconnect ------")
+			_, err := client.FetchJWTToken()
+			if err != nil {
+				log.Errorf("dbusEventLoop: error fetching JWT token")
 			}
-			log.Tracef("(dbusEventLoop) posting Event: %s", e.event)
-			d.serverUrl = serverURL
-			d.serverJwt = jwtToken
-			d.postEvent(e)
-			needsReconnect = false
 		}
 	}
-
 	log.Trace("dbusEventLoop: returning")
 }
 
--- a/app/daemon_test.go
+++ b/app/daemon_test.go
@@ -739,94 +739,113 @@ func TestTimeToSweepSessions(t *testing.T) {
 func TestDBusEventLoop(t *testing.T) {
 	currentUser, err := user.Current()
 	if err != nil {
-		t.Errorf("cant get current user: %s", err.Error())
-		return
+		t.Fatalf("cant get current user: %s", err)
 	}
 
 	testCases := []struct {
 		name    string
-		err     error
 		token   string
+		err     error
 		timeout time.Duration
 	}{
 		{
-			name:  "stopped-gracefully",
-			token: "the-token",
+			name:    "stopped-gracefully",
+			token:   "the-token",
+			timeout: 4 * time.Second,
 		},
 		{
 			name:    "token_not_returned_wait_forever",
 			token:   "",
 			timeout: 15 * time.Second,
 		},
+		{
+			name:    "token_not_returned_try_reconnection",
+			token:   "",
+			timeout: 2 * time.Second,
+		},
 	}
 
 	for _, tc := range testCases {
-		if tc.name == "token_not_returned_wait_forever" {
-			timeout := time.After(tc.timeout)
-			done := make(chan bool)
-			go func() {
-				t.Run(tc.name, func(t *testing.T) {
-					d := NewDaemon(&config.MenderShellConfig{
-						MenderShellConfigFromFile: config.MenderShellConfigFromFile{
-							ShellCommand: "/bin/sh",
-							User:         currentUser.Name,
-							Terminal: config.TerminalConfig{
-								Width:  24,
-								Height: 80,
-							},
-						},
-					})
+		d := NewDaemon(&config.MenderShellConfig{
+			MenderShellConfigFromFile: config.MenderShellConfigFromFile{
+				ShellCommand: "/bin/sh",
+				User:         currentUser.Name,
+				Terminal: config.TerminalConfig{
+					Width:  24,
+					Height: 80,
+				},
+			},
+		})
 
-					dbusAPI := &dbusmocks.DBusAPI{}
-					defer dbusAPI.AssertExpectations(t)
-					client := &authmocks.AuthClient{}
-					client.On("WaitForJwtTokenStateChange").Return([]dbus.SignalParams{
-						{
-							ParamType: "s",
-							ParamData: tc.token,
-						},
-					}, tc.err)
-					client.On("GetJWTToken").Return(tc.token, "", tc.err)
-					d.dbusEventLoop(client)
-				})
-				done <- true
-			}()
+		dbusAPI := &dbusmocks.DBusAPI{}
+		defer dbusAPI.AssertExpectations(t)
+		client := &authmocks.AuthClient{}
 
-			select {
-			case <-timeout:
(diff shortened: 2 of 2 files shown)


app/daemon_test.go
```diff
@@ -739,94 +739,120 @@ func TestTimeToSweepSessions(t *testing.T) {
...
+			dbusAPI := &dbusmocks.DBusAPI{}
+			defer dbusAPI.AssertExpectations(t)
+			client := &authmocks.AuthClient{}
+
+			switch tc.name {
+			case "token_not_returned_wait_forever":
+				timeout := time.After(tc.timeout)
+				done := make(chan bool)
+				go func() {
+					t.Run(tc.name, func(t *testing.T) {
```
lluiscampos, line 790: I don't get why we have `t.Run(...)` inside another `t.Run(...)`, can you explain me the intention?

app/daemon_test.go
```diff
@@ -739,94 +739,120 @@ func TestTimeToSweepSessions(t *testing.T) {
 func TestDBusEventLoop(t *testing.T) {
 	currentUser, err := user.Current()
 	if err != nil {
-		t.Errorf("cant get current user: %s", err.Error())
-		return
+		t.Fatalf("cant get current user: %s", err)
```
lluiscampos, line 742: Nice improvement :+1:

app/daemon_test.go
```diff
@@ -739,94 +739,120 @@ func TestTimeToSweepSessions(t *testing.T) {
...
+						client.AssertCalled(t, "FetchJWTToken")
+					})
+					done <- true
 				}()
-				d.dbusEventLoop(client)
-			})
-		}
+				select {
+				case <-timeout:
+					t.Logf("ok: expected to run forever")
```
lluiscampos, line 851: Nitpick (here and above): use `Log()` when there is no formatting involved.

lluiscampos (changes requested):
The fix looks good to me. Have you experimented a bit with the fix on a QEMU run? I think this deserves some manual testing also on top of our tests, for good measure.

See my comments below.

One other thing: the changelog deserves a bit more info. I would keep the commit title short, and write a longer entry that describes more of the historical reason for this. For example:

Fix bla bla bla.
Previously, `mender-connect` relied the Mender client to request new JWT tokens. Since the split of `mender` into `mender-update` and `mender-auth`, it is theoretically possible to run `mender-connect` without `mender-update`, in which case it wouldn't request JWT tokens without this fix.

app/daemon_test.go
SebOpsahl: Oh it really isn't needed, so changed it now. Good catch, thanks :)

SebOpsahl:
> The fix looks good to me. Have you experimented a bit with the fix on a QEMU run? I think this deserves some manual testing also on top of our tests, for good measure.

No, I am not aware on how to experiment with it on a QEMU run. But I have manually tested it with a build with yocto and mender-integration by watching how it reacts in a live build.

SebOpsahl:
I fixed the requests you made, so it should look good now. Good catches, thanks :)

lluiscampos (approved):

SebOpsahl:

SebOpsahl:
When running the big test the “test:acceptance:qemux86_64:uefi_grub:cross-platform” test fails, even though “test:acceptance:qemux86_64:uefi_grub” does not. So some of the cross-platform tests fail and returns the following errors, which seems to be due to loss of connection.
- FAILED test_mender-connect.py::TestMenderConnect::test_mender_connect_auth_changes - Failed: Timed out waiting for 'Connection established with http://localhost:5000/' (URL to view test: https://github.com/mendersoftware/meta-mender/blob/2ec7f2e0e83c6557fd1b8ad74f88c733c3bf28c6/tests/acceptance/test_mender-connect.py#L274) 

- FAILED test_mender-connect.py::TestMenderConnect::test_mender_connect_reconnect - Failed: Timed out waiting for 'connection manager failed to connect' (URL to view test: https://github.com/mendersoftware/meta-mender/blob/2ec7f2e0e83c6557fd1b8ad74f88c733c3bf28c6/tests/acceptance/test_mender-connect.py#L274)

How may it be that it works for the other tests, but not testing for cross platform?

lluiscampos:
> (...) How may it be that it works for the other tests, but not testing for cross platform?

The test is only run for that platform. If you look into the [details of the test](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1201262197/test_report?job_name=test%3Aacceptance%3Aqemux86_64%3Abios_grub) report you will find them mark as `skip`.

The name might be confusing, but the "cross platform" tests are the ones that are independent of the platform so we run them only once. The tests at hand are runtime tests for the mender-connect software, so these we consider "cross platform" because it doesn't matter how you integrate Mender on the platform, the software will run the same way once on the device. (Does this explanation make sense?)

We need to figure out if the tests are failing due to a failure in the software or is the test that needs some modification. For example if we have modified the text in the logs or the order of them it is possible that the test expectations are not met but the software is still fine.

SebOpsahl:
Oh okey, I understand now. Will look into it :+1:

SebOpsahl:

SebOpsahl:

SebOpsahl:
Problem to fix:
Make `mender-connect` request a new JWT token on its own, so its not dependent on `mender-update` to do so.

Proposed solution (PR):
Previously **FetchJWTToken()** wasn’t called which wouldn’t let it call the JWT token independently, which made it dependent on **mender-update**. The PR proposes a solution where **FetchJWTToken()** is used to fetch the token, to then for **WaitForJwtTokenStateChange()** to retrieve it in it’s waiting. The solution should look good.

The error:
When running the big test some of the **test:acceptance:qemux86_64:uefi_grub:cross-platform** tests fail. The error message returned is:
- FAILED test_mender-connect.py::TestMenderConnect::test_mender_connect_auth_changes - Failed: Timed out waiting for 'Connection established with http://localhost:5000/' (URL to view test: https://github.com/mendersoftware/meta-mender/blob/2ec7f2e0e83c6557fd1b8ad74f88c733c3bf28c6/tests/acceptance/test_mender-connect.py#L274) 

- FAILED test_mender-connect.py::TestMenderConnect::test_mender_connect_reconnect - Failed: Timed out waiting for 'connection manager failed to connect' (URL to view test: https://github.com/mendersoftware/meta-mender/blob/2ec7f2e0e83c6557fd1b8ad74f88c733c3bf28c6/tests/acceptance/test_mender-connect.py#L274)
Which both are connection related errors. The one place one of there errors comes from is here 
```python
def test_mender_connect_auth_changes(
        self, request, connection, with_mock_files, with_mock_servers,
    ):
        """Test that mender-connect can re-establish the connection on D-Bus signals"""

        try:
            dbus_set_token_and_url(connection, "token1", "http://localhost:5000")

            # start the mender-connect service
            startup_time = qemu_system_time(connection)
            connection.run(
                "systemctl --job-mode=ignore-dependencies start mender-connect"
            )

            # wait for first connect
            _ = wait_for_string_in_log(
                connection,
                startup_time,
                30,
                "Connection established with http://localhost:5000",
            )
```

Proposed change in test:
The proposed change for the test. Is to set the token, wait for a connection and then send a signal, simulating the task of **FetchJWTToken()**
```python
dbus_set_token_and_url(connection, "token1", "http://localhost:5000")

# start the mender-connect service
connection.run(
"systemctl --job-mode=ignore-dependencies start mender-connect"
)

# wait to secure connection
time.sleep(100)
dbus_emit_signal(connection)
signal_time = qemu_system_time(connection)

# wait for first connect
_ = wait_for_string_in_log(
connection,
signal_time,
30,
"Connection established with http://localhost:5000",
)
```
The current focus is towards changing the mocks for the test function, to better fit the new implementation. Since the solution should look good

SebOpsahl:
@kacf Problem to fix:
Make `mender-connect` request a new JWT token on its own, so its not dependent on `mender-update` to do so.

Proposed solution (PR):
Previously **FetchJWTToken()** wasn’t called which wouldn’t let it call the JWT token independently, which made it dependent on **mender-update**. The PR proposes a solution where **FetchJWTToken()** is used to fetch the token, to then for **WaitForJwtTokenStateChange()** to retrieve it in it’s waiting. I and Lluis have talked an the solution should look good.

Proposed change in test:
Since the tests is based on mocks it is proposed to set the token, wait for a connection and then send a signal, simulating the task of **FetchJWTToken()**

With these changes the test_mender_connect_auth_changes() which previously didn’t work now is successful. [This test case](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/test_mender-connect.py#312) however from the test_mender_connect_reconnect() test function fails. All the other test cases from the function is successful. The one thing separating the failing test case from the others is that this test case focuses on killing the server and then waiting for the error.
Any idea as to what might be causing the error? The error returned is **Failed: Timed out waiting for 'error reconnecting:'**

app/daemon.go
```diff
@@ -312,57 +312,49 @@ func (d *MenderShellDaemon) dbusEventLoop(client mender.AuthClient) {
 			break
 		}
 
-		if d.needsReconnect() {
```
SebOpsahl, line 315: In this newest commit I removed this extra check "if d.needsReconnect" early in the function which set the needsReconnect value to True. Rather the code now only have one check. Having these two checks doing the same thing must have delayed the tests and made it fail.

app/daemon.go
```diff
@@ -312,57 +312,49 @@ func (d *MenderShellDaemon) dbusEventLoop(client mender.AuthClient) {
...
+					d.postEvent(e)
+					log.Tracef("(dbusEventLoop) posting Event: %s", e.event)
+					needsReconnect = false
 
 					// If the server (Mender client proxy) closed the connection, it is likely that
 					// both messageLoop is asking for a reconnection and we got JwtTokenStateChange
 					// signal. So drain here the reconnect channel and reconnect only once
 					if d.needsReconnect() {
-						log.Debug("dbusEventLoop: drained reconnect req channel")
+						log.Trace("dbusEventLoop: daemon needs to reconnect")
```
SebOpsahl, line 344: Reference point for message above. Only now one check for setting **needsReconnect** to **True**

app/daemon.go
```diff
@@ -312,57 +312,49 @@ func (d *MenderShellDaemon) dbusEventLoop(client mender.AuthClient) {
...
-					needsReconnect = true
+					e := MenderShellDaemonEvent{
+						event: EventReconnect,
+						data:  token,
+						id:    "(dbusEventLoop)",
+					}
+					d.serverJwt = token
+					d.serverUrl = serverURL
+					d.postEvent(e)
+					log.Tracef("(dbusEventLoop) posting Event: %s", e.event)
```
lluiscampos, line 337: Either we move this entry to _before_ the `d.postEvent(e)` line or we change the text for something like "posted Event: ..."

app/daemon.go
lluiscampos: Okey. But I think now the comment above with regards to "draining" the channel is obsolete, as with the current code this is the only point where the message loop can signal a re-connection.

I think the code is good but we need to adjust the miss-leading comment (or remove?)

lluiscampos (approved):

alfrunes:
This PR looks stale, @lluiscampos can you take a look and see if we can merge it down?

lluiscampos:
@alfrunes Unfortunately, it requires quite a bit of work with the tests:
* https://github.com/mendersoftware/meta-mender/pull/2108

If you prefer, I can close both PRs and reference them from the JIRA task for the future.

alfrunes:
> @alfrunes Unfortunately, it requires quite a bit of work with the tests:
> 
> * [chore: updating mocks for tests to test mender-connect JWT request handling meta-mender#2108](https://github.com/mendersoftware/meta-mender/pull/2108)
> 
> If you prefer, I can close both PRs and reference them from the JIRA task for the future.

Ahh, I see. It looked like it was just waiting for an owner to merge. Then let's close and add a reference it in the ticket if nobody is currently owning it.

PR #138, opened 2024-08-01
fix: systemd: Always try restarting the client if it exits.

MuchoLucho:
Just applying the same logic as 857bae996f544a6abc67ec6ca748787b2e98e604 from the mender client repository. It can prevent the service not to be triggered due to a corner case.

Changelog: Title
Ticket: None

### DIFF
--- a/support/mender-connect.service
+++ b/support/mender-connect.service
@@ -8,7 +8,7 @@ Type=idle
 User=root
 Group=root
 ExecStart=/usr/bin/mender-connect daemon
-Restart=on-abort
+Restart=always
 
 [Install]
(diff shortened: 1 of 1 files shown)


lluiscampos:
I don't really understand the difference, from systemd point of view. I am pinging @vpodzime and @TheYoctoJester for their advise :smiley:

tranchitella (approved):

tranchitella:
> I don't really understand the difference, from systemd point of view. I am pinging @vpodzime and @TheYoctoJester for their advise 😃

If the service exits with an error, restart won't actually start it.

lluiscampos:

PR #157, opened 2025-07-02
feat: lower backoff intervals

danielskinstad:
The previous backoff had a starting interval at 1
minute and a max backoff at 60 minutes. When it reached 60 minutes it
would linearly increase to 120 minutes minutes where it would keep
retrying.

The backoff will now start at 1 second and exponentially increase after
3 retries until it reaches 30 minutes where it will keep retrying.
For each interval a jitter between 0 and 5 seconds will be added.

Ticket: ME-546
Changelog: Commit

### DIFF
--- a/connectionmanager/connectionmanager.go
+++ b/connectionmanager/connectionmanager.go
@@ -66,11 +66,10 @@ func GetWriteTimeout() time.Duration {
 }
 
 var a *expBackoff = &expBackoff{
-	maxInterval:  60 * time.Minute,
-	attempts:     0,
-	exceededMax:  false,
-	maxBackoff:   120 * time.Minute,
-	smallestUnit: time.Minute,
+	attempts: 0,
+	// Set max backoff to 30 minutes
+	maxBackoff:   1800 * time.Second,
+	smallestUnit: time.Second,
 }
 
 func connect(
--- a/connectionmanager/connectionmanager_test.go
+++ b/connectionmanager/connectionmanager_test.go
@@ -138,8 +138,6 @@ func TestConnectFailed(t *testing.T) {
 	old := a
 	a = &expBackoff{
 		attempts:     0,
-		maxInterval:  60 * time.Second,
-		exceededMax:  false,
 		maxBackoff:   120 * time.Second,
 		smallestUnit: time.Second,
 	}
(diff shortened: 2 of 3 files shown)


connectionmanager/connectionmanager_test.go
```diff
@@ -173,6 +170,15 @@ func TestExponentialBackoffTimeCalculation(t *testing.T) {
 		intvl := b.GetExponentialBackoffTime()
 		assert.Equal(t, intvl, 1*time.Minute)
 	}
+	// Test with 1 minute interval + randomInterval
+	for i := 0; i < 3; i++ {
+		b.attempts = i
+		b.randomInterval = 5*time.Second
```
elkoniu, line 176: If we decided to use `smallestUnit` I would stick to that and get rid of `time.Seconds` in favor of `smallestUnit`. We can keep the comment explaining that it most likely be 5 seconds.

connectionmanager/connectionmanager_test.go
```diff
@@ -173,6 +170,15 @@ func TestExponentialBackoffTimeCalculation(t *testing.T) {
...
+		intvl := b.GetExponentialBackoffTime()
+		assert.Equal(t, intvl, (1*time.Minute)+b.randomInterval)
```
elkoniu, line 178: Same here - if we are using `time.Minute` and `time.Second` what is the point of `smallestUnit` ?

elkoniu (commented):
Generally LGTM, just few minor questions :)

connectionmanager/connectionmanager_test.go
danielskinstad: Removed this test and added the random interval to where it was needed

connectionmanager/connectionmanager_test.go
danielskinstad: I didn't see that in the tests, changed it to smallestUnit :)

connectionmanager/connectionmanager_test.go
```diff
@@ -163,61 +161,51 @@ func TestWriteFailed(t *testing.T) {
...
-		maxInterval:  60 * time.Minute,
-		exceededMax:  true,
 		maxBackoff:   120 * time.Minute,
 		smallestUnit: time.Minute,
 	}
 	for i := 0; i < 10; i++ {
 		intvl := b.GetExponentialBackoffTime()
-		assert.GreaterOrEqual(t, intvl, time.Duration(120*time.Minute))
-		assert.LessOrEqual(t, intvl, time.Duration(122*time.Minute))
+		assert.GreaterOrEqual(t, intvl, time.Duration(120*b.smallestUnit))
```
michalkopczan, line 207: Max backoff time is 30 minutes. Here we check for 120-122. We initialize the struct to 120 minutes maxbackoff.
I know we test only the GetExponentialBackOffTime here, but it can be confusing.
Maybe make the maxBackoff in connectionmanager.go a const, and reference it in a tests as a default?

danielskinstad: I don't think it really matters that much as it's just testing the logic, but I modified the tests and set the maximum interval to 30 minutes

danielskinstad:

connectionmanager/exponentialbackoff.go
```diff
@@ -25,53 +25,44 @@ import (
...
 
-	if nextInterval > a.maxInterval && !a.exceededMax {
-		a.exceededMax = true
-		a.attempts = 0
-	}
-
-	if a.exceededMax {
-		return a.maxInterval + (time.Duration(a.attempts) * time.Minute) + randomInterval
+	if nextInterval > a.maxBackoff {
+		return a.maxBackoff + a.randomInterval
```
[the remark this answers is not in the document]
elkoniu: ACK

vpodzime (approved):
Looks good to me either way.

connectionmanager/connectionmanager.go
```diff
@@ -66,11 +66,10 @@ func GetWriteTimeout() time.Duration {
...
 
 var a *expBackoff = &expBackoff{
-	maxInterval:  60 * time.Minute,
-	attempts:     0,
-	exceededMax:  false,
-	maxBackoff:   120 * time.Minute,
-	smallestUnit: time.Minute,
+	attempts: 0,
+	// Set max backoff to 30 minutes
+	maxBackoff:   1800 * time.Second,
```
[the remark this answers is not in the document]
danielskinstad: Saw your comment _after_ I merged, It could be 30 minutes, yes, but it doesn't matter that much as this value isn't exposed outside of the code + I added the comment so I think it's fine,

PR #158, opened 2025-07-09
feat: enable 10 minute idle timeout by default 

danielskinstad:
Introduce a default value of 10 minutes for `ExpireAfterIdle`. This default applies when `StopExpired` is true and both `ExpireAfter` and `ExpireAfterIdle` are unset, since these two options are mutually exclusive.

Note that this is a behavioral change; no default value was previously assigned to `ExpireAfterIdle`.

Changelog: Commit
Ticket: MEN-8260

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -460,3 +460,72 @@ func TestServerArgumentsDeprecated(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Contains(t, buf.String(), "ServerCertificate field is deprecated")
 }
+
+func TestConfigurationDisableSessionExpirationSettings(t *testing.T) {
+	var configJSON = `{
+		"Sessions": {
+			"StopExpired": false,
+			"ExpireAfter": 40,
+			"ExpireAfterIdle": 40
+		}
+	}`
+
+	configFile, err := os.Create("main.config")
+	assert.NoError(t, err)
+	defer os.Remove("main.config")
+	configFile.WriteString(configJSON)
+
+	config, err := LoadConfig("main.config", "")
+	assert.NoError(t, err)
+	assert.NotNil(t, config)
+	config.Validate()
+
+	assert.Equal(t, false, config.Sessions.StopExpired)
+	assert.Equal(t, uint32(0), config.Sessions.ExpireAfter)
+	assert.Equal(t, uint32(0), config.Sessions.ExpireAfterIdle)
+}
+
+func TestConfigurationDefaultSessionExpirationSettings(t *testing.T) {
+	var configJSON = `{
+		"ShellCommand": "/bin/bash",
+		"ReconnectIntervalSeconds": 123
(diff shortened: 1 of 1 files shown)


TheMeaningfulEngineer:
> This default applies when StopExpired is true and both ExpireAfter and ExpireAfterIdle are unset, since these two options are mutually exclusive.

Want to over-communicate so we're on the same page.
Sorry if obvious and I'm miss-interpreting.

Reading this is looks like user will still need to set `StopExpired` to get the default behavior.
If so, my bad in not making in clear in the ticket.

The end goal is to introduce a default such that if a user installs mender-connect and doesn't modify the configuration session it closes after 10 minutes of idle, whatever that translates to in the config.

If the user needs to set a `Stop expired` to enabled in the config to get the 10min timeout on idle, that misses the point.

TheMeaningfulEngineer:
> Right, I get your point. The only function of StopExpired is Whether to stop expired sessions, and since this is false by default it will (and should I guess) disable both ExpireAfter and ExpireAderIdle. To get the desired behavior we would need to set StopExpired to true by default, which I guess is fine

Yes, this is the default behavior.
So both please.
Sorry on the wrong AC in the ticket.

config/config.go
```diff
@@ -50,7 +52,7 @@ type PortForwardConfig struct {
 
 type SessionsConfig struct {
 	// Whether to stop expired sessions
-	StopExpired bool
+	StopExpired *bool
```
vpodzime, line 55: Why this change? Can't we simply assign the boolean value of `DefaultStopExpired` below?

vpodzime (commented):
:heavy_plus_sign: *timout* in the commit title. Looks good to me otherwise.

config/config.go
danielskinstad: The refact touched on too much and I didn't bother. I ended up setting the default value when initializing the struct, I think that's sufficient.

config/config.go
```diff
@@ -271,7 +273,13 @@ func (c *MenderShellConfig) applyDefaults() error {
 		c.Sessions.ExpireAfterIdle = 0
 	} else {
 		if c.Sessions.ExpireAfter > 0 && c.Sessions.ExpireAfterIdle > 0 {
-			log.Warnf("both ExpireAfter and ExpireAfterIdle specified.")
+			log.Warnf("Both ExpireAfter and ExpireAfterIdle specified; " +
+				"they are mutually exclusive")
```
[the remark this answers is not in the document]
danielskinstad: I tried to add an error here but it made some unit tests fail (I didn't dig too deep into it TBH, maybe I'll look more tomorrow). I don't know if it's worth it adding an extra restriction that might break peoples configs?

config/config.go
vpodzime: Makes sense to me -- set the default and let it be overridden.

PR #170, opened 2026-02-20
ci: add qa-common retry and runner defaults

oldgiova:
Include qa-common/runner.yml and qa-common/retry.yml from mendertesting to apply a consistent .qa-common-default-runner tag and .qa-common-default-retry policy to all jobs pipeline-wide.

Add network probes (apt/go/git-clone) to jobs that perform network operations, so transient failures trigger a retry instead of a hard fail.

Ticket: QA-1490


vpodzime (approved):
Looks good to me.

oldgiova:
non functional change; medner-qa pipeline not needed here

PR #174, opened 2026-05-07
MEN-5902: implement port-forwarding protocol v2

alfrunes:
The new protocol is identical to v1 except v2 does not stop and wait for
message acknowledgement. This immensely speeds up the throughput
especially for higher latency. `mender-connect` continues to support the
previous version of port-forwarding for backward compatibility.
Requires mender-server > 4.1 and mender-cli > 2.0.


alfrunes:

session/portforward.go
```diff
@@ -332,19 +363,19 @@ func (h *PortForwardHandler) portForwardHandlerForward(
...
 		err := portForwarder.Write(message.Body)
-		// send ack
-		response := &ws.ProtoMsg{
-			Header: ws.ProtoHdr{
-				Proto:     message.Header.Proto,
-				MsgType:   wspf.MessageTypePortForwardAck,
-				SessionID: message.Header.SessionID,
-				Properties: map[string]interface{}{
-					wspf.PropertyConnectionID: connectionID,
+		if h.proto == ws.ProtoTypePortForward && err == nil {
```
danielskinstad: Previously it sent ack regardless of the error. Will this change the behavior of v1?

alfrunes: Interesting. You are right, but the error from the write is completely ignored in the previous implementation :thinking:

alfrunes: I think I would consider this a bug in the previous implementation. It does not make sense to proceed if the message was not actually forwarded :shrug:

danielskinstad: Alright. Do you think it's worth adding the error check in a separate commit and backporting it?

danielskinstad (approved):

session/portforward.go
alfrunes: I put the changes in here: 528d50c35b31e02754e3610a08746d5a33f94995
While doing that I also added another fix: 95bb63f9845ce55c07d5af425c1d91948dfa6eff
Can you please have another look?

danielskinstad (approved):
LGTM

PR #182, opened 2026-06-26
ci: Use test:unit from mendertesting template

elkoniu:
mender-connect had a local implementation of golang coverage test. Instead the one from the template shall be used - the same way as in other repositories.


elkoniu:

elkoniu:
Closed, to be tracked in QA-1684 as most likely it will require template bump in `mendertesting`

PR #198, opened 2026-08-13
fix: Do not block the message loop when a late pong arrives

otomon10:
## Problem

A single remote-terminal session whose healthcheck times out can permanently hang the whole mender-connect daemon.

`MenderShellSession.pong` is an unbuffered channel whose only reader is the `select` in the `healthcheck` goroutine. When the healthcheck times out, that goroutine logs `health check failed`, calls `MenderShellStopById` and returns — from that point on the channel has no reader.

The session, however, stays reachable through `MenderShellSessionGetById` until `MenderShellStopById` finishes removing it. Removal takes at least four seconds, because `procps.TerminateAndWait` unconditionally runs `SIGTERM → sleep 2s → SIGKILL → sleep 2s`. So a normal healthcheck-timeout stop opens a removal window of at least four seconds in which a late pong finds the session and `HealthcheckPong()` blocks forever on the channel send.

`HealthcheckPong()` is called synchronously on the message-loop path (`messageLoop` → `routeMessage` → `routeMessagePongShell`), so the blocked send hangs the daemon's message loop: the daemon stops reading the websocket, goes silent in the logs, and never reconnects. The process stays alive, so systemd's `Restart=always` never fires. Remote terminal connections from the Mender UI then fail (`WebSocket error: Unknown error` in the Troubleshooting tab) until mender-connect is restarted manually.

We hit this in production — remote access to a field device was lost for three days until a manual restart — and we can reproduce it on demand on a real device against hosted Mender by delaying a single pong so that it arrives inside the removal window. Goroutine dump of the hung daemon (from a 2.3.2-based binary; line numbers on current master differ only in the two `daemon.go` frames):

```
goroutine 36 [chan send, 4 minutes]:
runtime.chansend1(...)
github.com/mendersoftware/mender-connect/session.(*MenderShellSession).HealthcheckPong(...)
	github.com/mendersoftware/mender-connect/session/shell.go:447
github.com/mendersoftware/mender-connect/app.(*MenderShellDaemon).routeMessagePongShell(...)
	github.com/mendersoftware/mender-connect/app/daemon_shell.go:264
github.com/mendersoftware/mender-connect/app.(*MenderShellDaemon).routeMessage(...)
	github.com/mendersoftware/mender-connect/app/daemon.go:579
github.com/mendersoftware/mender-connect/app.(*MenderShellDaemon).messageLoop(...)
	github.com/mendersoftware/mender-connect/app/daemon.go:256
```

The dump also shows that the `healthcheck` goroutine that would read the pong no longer exists, so this send can never complete.

## Affected versions

All versions since 1.1.0 (April 2021) are affected, up to and including current master: the unbuffered send is unchanged across this range, and we reproduced the blocked send on eleven tags from 1.1.0 through master.

## Fix

Give `pong` a single slot and make the send non-blocking. A pong only signals that the peer is alive, so holding one is enough and dropping a second one changes nothing. The buffer keeps a pong that arrives while the healthcheck is momentarily outside its `select`; the `default` keeps the sender from blocking once the reader is gone.

## Testing

- New `session/shell_test.go`:
  - `TestLatePongDuringHealthcheckStop` delivers a pong right after `health check failed` is logged, i.e. inside the removal window. Without the fix it fails after 30 s with the sending goroutine blocked at `session/shell.go:447`; with the fix it passes. The trigger is event-driven (a log hook), not sleep-based, and the window it relies on is at least four seconds wide, so it is not sensitive to scheduling.
  - `TestPongAfterSessionDeleted` (negative control): a pong delivered after the removal is rejected with `ErrSessionNotFound` and never blocks.
- The two tests add roughly 12 s of wall time to the suite.
- Note: `go test -race ./session` fails on master even without this patch (pre-existing data races); CI runs without `-race`.
- Verified on a real device (a 2.3.2-based binary with this patch, against hosted Mender): the patched binary survives the same late-pong injection that reliably hangs the unpatched one, and keeps accepting new terminal sessions afterwards.

We ship mender-connect from the 2.3.x series, so we would appreciate a backport to the maintained release branches. The patch applies cleanly there: `session/shell.go` is the same blob on master, 2.3.x and 3.0.x.

### DIFF
--- a/session/shell.go
+++ b/session/shell.go
@@ -161,7 +161,7 @@ func NewMenderShellSession(
 		sessionType: ShellInteractiveSession,
 		status:      NewSession,
 		stop:        make(chan struct{}),
-		pong:        make(chan struct{}),
+		pong:        make(chan struct{}, 1),
 	}
 	sessionsMap[sessionId] = s
 	sessionsByUserIdMap[userId] = append(sessionsByUserIdMap[userId], s)
@@ -444,7 +444,14 @@ func (s *MenderShellSession) healthcheckPing() {
 }
 
 func (s *MenderShellSession) HealthcheckPong() {
-	s.pong <- struct{}{}
+	// The healthcheck goroutine stops reading this channel once it takes its
+	// timeout branch, but the session can still be looked up for a while, so
+	// a blocking send here could leave the daemon's messageLoop waiting
+	// forever on a late pong.
+	select {
+	case s.pong <- struct{}{}:
+	default:
+	}
 }
 
 func (s *MenderShellSession) ShellCommand(m *ws.ProtoMsg) error {
--- a/session/shell_test.go
+++ b/session/shell_test.go
@@ -1,0 +1,211 @@
+// Copyright 2026 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package session
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"os/user"
+	"runtime"
+	"strconv"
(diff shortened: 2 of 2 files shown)


alfrunes (approved):
This change looks good to me.

Just as a side-note, there seems to be lacking proper cleanup when the healthcheck go routine times out. Instead the go routine just silently shuts down after signalling the session is shutting down leading to the race to begin with. I'm noting this down as a follow-up :memo:

Thank you for fixing the most crucial part of the problem! 🙏

alfrunes:

